PR #8, opened 2016-05-03
Add monolog breadcrumb support

dcramer:
This doesn't work yet. No idea why. About to board a flight and will mess around with it when I have network again.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "require": {
         "php": ">=5.2.4",
         "illuminate/support": "4.*|5.*",
-        "sentry/sentry": "*"
+        "sentry/sentry": ">=0.16.0"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


dcramer:
Oh this will also require changing the sentry/sentry requirement to whatever version we pin it to that has breadcrumbs

dcramer:
Should fully work (tested in local). Needs 0.16.x of sentry/sentry

dcramer:
This might not yet work with Laravel 4.2 from a quick test locally.

dcramer:
Laravel 4 has same issue as Ruby (getsentry/raven-ruby#497)

- need to parse errors out of logs
- need to prevent dupe final log entry

screenshot 2016-05-03 14 01 29

dcramer:
We still need to deal with dupes, but thats a larger issue.

PR #10, opened 2016-05-05
Remove logging message

Stayallive:
I see the was to plan to remove already but has not been done?

This is spamming my server logs pretty bad actually :(


dcramer:
@Stayallive is there a better practice for this? We do it in other langs, but I guess its noisy in PHP since it doesn't have long lived processes.

I'll accept this for now, but we should consider if theres other things we can/should do to allow debug info like this.

Stayallive:
I have honestly never seen this done in any package before.

Is this to just make sure it is visible that Sentry is enabled and sending?

PR #15, opened 2016-05-18
Add /examples to export-ignore

dcramer:
According to the internet, this might stop ``php artisan optimize`` from including files we don't want to be part of the build.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/examples export-ignore


stayallive:
I believe this only works if composer is set to favor dist (since github uses this ignore to prevent the files ending up in a tarball or zipball) instead of cloning each dependency.

I don't see however how `php artisan optimize` will include these files... where on the internet is this stated? Sounds interesting.

dcramer:
@stayallive its unclear how its happening to me as well -- it was a support ticket from getsentry.com

stayallive:
Hmm. Weird. I checked but it is not a problem for me. But this change is nice anyhow just for the extra kb's saved 😎

PR #17, opened 2016-05-29
Remove wrong import from middleware context example

stayallive:
I just noticed I left this import in my example code (A)


dcramer:
thanks!

PR #26, opened 2016-08-01
Cleanup

jbrooksuk:
- Laravel has a min PHP constraint on 5.5.9, so we're now matching that.
- Since we're using at least PHP 5.4 we're now using short arrays everywhere. There was already one array using this syntax (so the code was immediately incompatible with 5.2.4 anyway).
- I've imported all namespaces, rather than using the lazy syntax. Also, we're importing the `Log` facade. It's possible that someone would remove the `Log` alias from `config/app.php` so this import would fail.

I think next I'll switch to PSR-4 (which I think can just be a `composer.json` modification) and then write some tests, if I get enough time.

---

This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-laravel/26)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 /examples export-ignore
+/test export-ignore
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": ">=5.2.4",
+        "php": ">=5.5.9",
         "illuminate/support": "4.*|5.*",
         "sentry/sentry": ">=0.16.0"
     },
--- a/src/Sentry/SentryLaravel/SentryFacade.php
+++ b/src/Sentry/SentryLaravel/SentryFacade.php
@@ -6,6 +6,11 @@ use Illuminate\Support\Facades\Facade;
 
 class SentryFacade extends Facade
 {
+    /**
+     * Get the registered name of the component.
+     *
+     * @return string
+     */
     protected static function getFacadeAccessor()
     {
         return 'sentry';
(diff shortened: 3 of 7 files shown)


stayallive:
We should match `php >=5.3.0` since that is what Laravel 4 requires which is supported by this library. This also means the short array syntax cannot be used yet 😢

jbrooksuk:
> We should match php >=5.3.0 since that is what Laravel 4 requires which is supported by this library. This also means the short array syntax cannot be used yet 😢

Hmm, good point. I didn't think about that. I wonder if @dcramer would be up for versioning the package? That way we can deprecate v4 support in one version.

stayallive:
@jbrooksuk doesn't that add unneeded complexity when changing things? There is so little code in here I see no reason to branch it all for code the user doesn't even touch or need to interact with. The underlying library `sentry/sentry-php` supports PHP 5.2 so there is no use really IMHO. 

However on that note it is a valid point to somewhere down the line start pushing to only support _not_ end-of-life PHP versions in newer versions of the `sentry/sentry-php` and slowly work towards that. The users that want or need to stay in the year __2006__ can use older versions still if they so require.

After `sentry/sentry-php` is requiring higher PHP versions it would be nice to "upgrade" this package too!

dcramer:
We generally support things within the same lib until the usage is very small. So if say if we have an issue on laravel 4 it's a bug in this library as the current version should fully support it. 

Generally on board with improving things, through as @statallive mentioned it adds currently unneeded complexity to maintain two libraries just so we can use some syntax niceties.

stayallive:
Closing this because it's old with a lot of conflicts but most things could be done now since we dropped Laravel 4.x support from version `0.9.x` onwards.

Feel free to reopen ofcourse if you want to 👍

PR #31, opened 2016-09-27
Improve breadcrumbs

dcramer:
- capture illuminate queries
- swap monolog for illuminate log

This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-laravel/31)

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,9 @@ return array(
 
     // capture release as git sha
     // 'release' => trim(exec('git log --pretty="%h" -n1 HEAD')),
+
+    // Capture bindings on SQL queries
+    'breadcrumbs.sql_bindings' => true,
 );
 ```
 
--- a/examples/laravel-4.2/app/config/packages/sentry/sentry-laravel/config.php
+++ b/examples/laravel-4.2/app/config/packages/sentry/sentry-laravel/config.php
@@ -5,4 +5,7 @@ return array(
 
     // capture release as git sha
     // 'release' => trim(exec('git log --pretty="%h" -n1 HEAD')),
+
+    // Capture bindings on SQL queries
+    'breadcrumbs.sql_bindings' => false,
 );
--- a/examples/laravel-4.2/app/routes.php
+++ b/examples/laravel-4.2/app/routes.php
@@ -7,13 +7,14 @@
 |
 | Here is where you can register all of the routes for an application.
 | It's a breeze. Simply tell Laravel the URIs it should respond to
-| and give it the Closure to execute when that URI is requested.
+| and give it the controller to call when that URI is requested.
 |
 */
 
 function verifyCredentials()
 {
     Log::info('Verifying credentials');
+    $user = DB::table('migrations')->where('migration', 'a migration')->first();
     throw new Exception('No credentials passed!');
 }
 
--- a/examples/laravel-5.1/config/app.php
+++ b/examples/laravel-5.1/config/app.php
@@ -78,7 +78,7 @@ return [
     |
     */
 
-    'key' => env('APP_KEY', 'SomeRandomString'),
+    'key' => env('APP_KEY', 'xlhF31NeOlibJcoOW9tvZg7TkHcAZI3a'),
 
     'cipher' => 'AES-256-CBC',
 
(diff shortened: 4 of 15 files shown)


dcramer:
@taylorotwell part of me was hoping to blindly capture all events, but it seems in some cases the Event has a string defining it (which is great!) and others its a class name (which is less great). Have you considered requiring consistency in event names in a future version?

dcramer:
@stayallive this seem sane to you? seems good in both 4 and 5. I'm not sure if theres other situations or very common events we should capture here.

dcramer:
And finally, interestingly in Laravel 4 illuminate.log seems like it might be better for us than Monolog:

screenshot 2016-09-27 13 35 07

stayallive:
@dcramer Going to check this out in a few hours when I wake up :)

stayallive:
@dcramer if you would like you can do something like the debugbar does with capturing all the events although I'm not sure if usefull (https://github.com/barryvdh/laravel-debugbar/blob/master/src/DataCollector/EventCollector.php). The other collectors in the debugbar source are also very interesting and might give you some ideas on what to collect.

Few that I can think of would be usefull are:
- `Illuminate\Routing\Events\RouteMatched`: To see which route is hit
- `Illuminate\Mail\Events\MessageSending`: See which mails are dispatched
- `Illuminate\Queue\Events\*`: This might be a bad idea in <= 5.2 because there the queue worker process runs indefinitly so they amount of event buildup would be crazy (seeing 100+ breadcrumbs already on queue jobs in a 5.2).

I looked in this list: https://laravel.com/docs/5.1/events#framework-events. These are almost all changed to use classes in 5.2+ but might give you an idea on what's there and what the framework emits. You can find event classes quickly by looking in the `vendor/illuminate/*/Events/*` folders.

I personally would like to be able to configure the level or verbosity these loggings happen because this could possibly be a overhead that is not worth it on every environment. I could imagine this could lead to memory problems and extremely large requests sent to Sentry with hundreds of breadcrumbs with large datasets (queries and or bindings).

Hope this is usefull 👍

dcramer:
@stayallive def makes sense -- maybe we'll bring along configuration similar to Laravel, in that you can specify either listeners to include or create a blocklist of ones you dont want. I will take a look at the debugbar implementation, as I agree some of these could be useful.

dcramer:
@stayallive only doing logs/queries for now, but i greatly improved it inspired by debugbar

stayallive:
@dcramer this looks very nice indeed 👍

dcramer:
Adding lumen support (if doable) and then gonna merge + release.

dcramer:
screenshot 2016-09-29 13 18 34

stayallive:
Lumen is using the same components for database and events as Laravel does so it should be very straight forward. 

Looking at that screenshot I already can see this is going to be a very usefull addition. Sometimes looking at data in objects where I can only see the classnames, with this I will be able to tell what I'm actually looking at 💯

dcramer:
@stayallive appears like the events may not exist by default in Lumen (at least struggling to get breadcrumbs showing up)

stayallive:
@dcramer do you have this uncommented? Might not register correctly otherwise. https://github.com/getsentry/sentry-laravel/blob/master/examples/lumen-5.2/bootstrap/app.php#L28

Although looking at the source, as long as events is bound all should work just fine. https://github.com/illuminate/database/blob/master/DatabaseManager.php#L189

Going to try get this working too... should be the same as Laravel I thought...

dcramer:
@stayallive nope but that doesnt seem to change it

stayallive:
Works for me, did you change your test config to actually log the queries (`config/sentry.php`)? 😜 

screen shot 2016-09-29 at 23 19 02

dcramer:
@stayallive well its just for logging bindings (queries will always log atm), but if it works for you I'm content :)

stayallive:
Yes even with Eloquent off it works like a charm 👍

dcramer:
This is now available in 0.4.0

stayallive:
Going to push this to prod right now, the transaction tracing also looks sweet!

PR #32, opened 2016-09-29
Add transaction name support

dcramer:
This uses the current route name (when possible) as the transaction name.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-laravel/32)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "require": {
         "php": ">=5.2.4",
         "illuminate/support": "4.*|5.*",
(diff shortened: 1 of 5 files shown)


dcramer:
screenshot 2016-09-29 15 39 00
screenshot 2016-09-29 15 39 20

dcramer:
I'm considering that we might want to split on ``\`` for some things as showing full class paths isn't overly useful here.

dcramer:
This is now available in 0.4.0

PR #33, opened 2016-09-30
Do not call boot method on Laravel Service Provider

stayallive:
The base Laravel serviceprovider does not have a boot method defined.


dcramer:
What version did this error on? (was following some docs I saw)

stayallive:
5.3

dcramer:
pushing 0.4.1

stayallive:
You are not required ever to do this since the boot method does nothing by itself it's just for registering things after everything has registered. But the failsafe for this is removed in 5.3. 

5.2 had this in it, removed in 5.3:
> https://github.com/laravel/framework/blob/c32d91c64504dd2baf40e8ba025ad348da4607a1/src/Illuminate/Support/ServiceProvider.php#L222-L238

PR #38, opened 2016-11-18
capture Laravel Log

rsands2801:


dcramer:
A couple of concerns here:

- This would need to be opt-in (not enabled by default)
- It needs to be configurable to some degree -- not everyone wants to send INFO events into Sentry (and to be honest, no one should be, as its next to useless)

rsands2801:
@dcramer let me work on making it configurable.  Also that's your opinion about it being useless - keep an open mind about what people use Sentry for ;)

dcramer:
@rsands2801 I only keep in mind what we build Sentry for, now what people decide to stuff in it :)

Logging generally serves a different purpose. We have very strong views on this and Sentry will continually be less and less useful for generic log-type messages. We wrote up a little bit of why we think thats so here: https://sentry.io/vs/logging/

rsands2801:
Article is interesting read but lets remember we capture the exceptions at the CI phase as well - so where does that leave Sentry....In fact if your not capturing exceptions in the test phase then your test phases are close to useless

AidasK:
I think this should supported too. Sentry won't get spammed if you will be sending only error or critical level log messages. 
It is really strange, then you write `\Log::error('Request failed')` and this error is not sent to sentry.

It also can be enabled by default, because framework itself doesn't generate these messages.

rsands2801:
@AidasK Definitely agree with your post.  I have finished the changes above and tests passed but it was never merged in so not sure what is happening with this request

dcramer:
We'll get this merged in, just didn't want to do it willy nilly. There's a lot of issues customers will have as soon as they start seeing these kinds of events (e.g. "why doesnt this group together?"), so I want to make sure we avoid as many as possible beforehand.

AidasK:
It would be great to have just a simple implementation. This feature can be disabled by default(as you said). If there are any issues it can be fixed. First implementation doesn't have to be perfect.

rsands2801:
@dcramer the issues will group together if the developer knows how to use the Laravel Log function correctly.

```
Log::error('My Error', ['custom_message' => 'error happening because...']);
```

will group with this in Sentry as it works with the initial value being the same as the exception name being thrown

```
Log::error('My Error', ['custom_message' => 'new error that has a different custom message will group...']);
```

rsands2801:
@dcramer can we look at merging this?  it is proving very painful without it

dcramer:
I will look at getting this in, but its still too opinionated by default.

- it shouldn't be setting tags (those are provided by other context)
- the user context should be already bound, so this is doing extra work
- the IP is part of the user context, and same as above, shouldn't be set on log events explicitly

rsands2801:
@dcramer I will just fork it and put up a package on packagist for it with the proper functionality.  Disappointing

rsands2801:
@dcramer is there a license issue or requirement for me doing so with a fork and creating as a package?

dcramer:
You can see our LICENSE in the root, but no there are no issues with you forking and hosting your own fork.

AidasK:
@dcramer Forking is not a solution. What is wrong with this pull request? Maybe we could work it out.

dcramer:
@AidasK I explained what needed to be done. See my comment above.

rsands2801:
@AidasK when I meet this level of resistance I am happy to just fork it as a solution.  The Laravel Log is not passing user context into sentry by default and I have no time to mess around making it work in some different way.  If Laravel Log doesn't pass it in then this commit allowed it to inject it in manually.  I am sure there is a fancier way of doing it but considering previous packages did it this way I simply left it like this.  If someone has more time to mess around with it, great - if not this will work with Laravel 5.* versions.

I am researching alternative companies that support Laravel properly and this fix is just an interim for me.

dcramer:
Feel free to open another PR with the appropriates changes, but this conversation is not constructive. We require certain principles to be met with code changes, starting with maintainability and safety for our users. We're not against these kinds of changes, but we're not going to just merge changes because of someone being excessively vocal.

Issue #24, opened 2016-07-11
Facade name collision with existing package


dcramer:
I'm trying to install Sentry on a Laravel 5 project but I'm already using Cartalyst Sentry for user authentication - which has an existing facade for "sentry." Is there a workaround for this?

dcramer:
My personal opinion is "this isn't our problem". I think you can use app('sentry') instead of the Facade, but maybe I'm wrong. Either way, I have no desire to cater to other people using the Sentry name for their own projects, especially since it dates back so far (and potentially we have trademark law on our side).

stayallive:
If it's not clashing in the IoC container (so `app('sentry')` is not the issue) people can simply change the facade name in their config files.

```php
    'Sentry' => Sentry\SentryLaravel\SentryFacade::class,
```

To:

```php
    'SentryTracker' => Sentry\SentryLaravel\SentryFacade::class,
```

Or something like that (you could also use `Raven` as facade / IoC name).

Although looking at their (depracated) code on GH it _is_ clashing in the IoC container. The only way around that is renaming Sentry in the IoC container (for this package or the cartalyst one 😉 ).

Possibly using `sentry.raven` is an solution since that would not clash with a package using `raven` as IoC binding I believe.

dcramer:
@stayallive thanks! fwiw we intend to completely remove "Raven" from the history books (too confusing to users)

stayallive:
@dcramer the using "Raven" is a terrible idea :P But no idea what would be nice to use instead then.

davzie:
I'm having this issue which is also stopping me using Sentry on an old Laravel 4 project that uses Cartalyst's package namespacing too. I actually took the service provider and botched it to use a new Facade called "Sentryerrors" instead, but alas this isn't working. Any idea how I can use my own provider to make this work? I love Sentry and desperately need error logging for this app :)

stayallive:
You should be able to copy the facade and service provider into your own application like this: https://gist.github.com/stayallive/2adc8a92aaa64ca5427f50141f715561

And add this Facade instead of the one in the docs:
```php
'SentryErrors' => App\Support\SentryFacade::class,
```

And add this ServiceProvider instead of the one in the docs:
```php
App\Support\SentryServiceProvider::class,
```

Also the config file is renamed to `sentry-laravel.php` in the gist above so you should rename the config file if it was already created.

Note: this code is untested and you might want to adjust the namespaces to suit you application.

davzie:
[...]

PR #39, opened 2016-11-30
Allow overriding abstract type Sentry is bound to in service container.

hackel:
This is a fix for #24.  I used a public static property so that you can also do this:
```php
class SentryFacade extends Facade
{
    protected static function getFacadeAccessor()
    {
        return SentryLaravelServiceProvider::$abstract;
    }
}
```


dcramer:
@hackel I'm going to merge this in, but what are your thoughts on making the default Facade be this type but titlecase?

hackel:
@dcramer I think that makes sense.  I tried to change as little as possible to make it work.  I don't generally use facades myself.  We could create the facade alias manually:
```php
if ($this->app['config']->get('auto_create_alias', true)) {
    \Illuminate\Foundation\AliasLoader::getInstance()->alias(studly_case(static::$abstract), SentryFacade::class);
}
```

Issue #40, opened 2016-12-09
Error with new breadcrumb feature


pedrommone:
I saw that now we can track breadcrumbs and thats pretty cool! But, unfortunately, one of our Laravel services have broke when we updated its dependencies. It was so painful that it even break the Exception Handler.

We removed the Sentry from the Exception Handler and got this log...

```
[2016-12-09 00:44:24] production.ERROR: Symfony\Component\Debug\Exception\FatalThrowableError: Type error: Too few arguments to function Illuminate\Auth\AuthManager::createDriver(), 0 passed in /data/www/vendor/laravel/framework/src/Illuminate/Support/Manager.php on line 87 and exactly 1 expected in /data/www/vendor/laravel/framework/src/Illuminate/Auth/AuthManager.php:16
Stack trace:
#0 /data/www/vendor/laravel/framework/src/Illuminate/Support/Manager.php(87): Illuminate\Auth\AuthManager->createDriver()
#1 /data/www/vendor/laravel/framework/src/Illuminate/Auth/AuthManager.php(18): Illuminate\Support\Manager->createDriver(NULL)
#2 /data/www/vendor/laravel/framework/src/Illuminate/Support/Manager.php(63): Illuminate\Auth\AuthManager->createDriver(NULL)
#3 /data/www/vendor/laravel/framework/src/Illuminate/Support/Manager.php(137): Illuminate\Support\Manager->driver()
#4 /data/www/vendor/sentry/sentry-laravel/src/Sentry/SentryLaravel/SentryLaravelServiceProvider.php(85): Illuminate\Support\Manager->__call('check', Array)
#5 /data/www/vendor/laravel/framework/src/Illuminate/Container/Container.php(738): Sentry\SentryLaravel\SentryLaravelServiceProvider->Sentry\SentryLaravel\{closure}(Object(Illuminate\Foundation\Application), Array)
#6 /data/www/vendor/laravel/framework/src/Illuminate/Container/Container.php(631): Illuminate\Container\Container->build(Object(Closure), Array)
#7 /data/www/vendor/laravel/framework/src/Illuminate/Foundation/Application.php(674): Illuminate\Container\Container->make('sentry', Array)
#8 /data/www/vendor/laravel/framework/src/Illuminate/Container/Container.php(1167): Illuminate\Foundation\Application->make('sentry')
[...]

PR #41, opened 2016-12-09
Gracefully handle errors capturing breadcrumbs

dcramer:
Refs GH-40

### DIFF
--- a/src/Sentry/SentryLaravel/SentryLaravelEventHandler.php
+++ b/src/Sentry/SentryLaravel/SentryLaravelEventHandler.php
@@ -26,8 +26,16 @@ class SentryLaravelEventHandler
 
     public function onWildcardEvent()
     {
-        $name = $this->events->firing();
         $args = func_get_args();
+        try {
+            $this->_onWildcardEvent($args);
+        } catch (\Exception $e) {
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


dcramer:
I'm not sure this is the best approach to logging these errors and its not something we yet do anywhere else in PHP.

stayallive:
I have been thinking about it and I think this is the only solution for that besides reporting it to Sentry itself. Loggin it through Laravel is also not possible since that would make it loop possibly.

But looking at #40 this would result in a log entry every single request in their app right?

pedrommone:
@stayallive Sentry only handle exception failures, right? If I'm right, its fine to `log` it back but would be nice if it send to Sentry too.

dcramer:
@pedrommone we pick up logs in certain situations, and also use them as breadcrumbs, which means it becomes complex and could be recursive

dcramer:
@stayallive yeah it'd hit repeatedly for #40, but our goal would also be to fix that -- this is mostly insurance

pedrommone:
@dcramer can you MR this one and open a new issue to figure out how to log without hit repeatedly?

dcramer:
@pedrommone you cannot avoid repeatedly logging -- they're unique requests with no global state

PR #42, opened 2016-12-09
Travis CI improvements 

pedrommone:
1. I've added the PHP 7.1
2. Why some versions are allowed to fail? I've removed them.


dcramer:
Super late on this, sorry! We allowed failures just to be aware, but at the end of the day we dont lookat it so no point in keeping them.

Issue #46, opened 2017-01-16
Laravel 5.4 compatibility


KKSzymanowski:
Laravel 5.4 will probably be released this or next month and there is a compatibility issue which will have to be addressed.

There has been some changes to `Illuminate\Events\Dispatcher`, eg. in [this](https://github.com/laravel/framework/commit/dbbfc62beff1625b0d45bbf39650d047555cf4fa) commit. It no longer exposes the `firing()` method used in `SentryLaravelEventHandler`. From what I understand it is used to retrieve the name of the current event and direct the control flow accordingly. 
In 5.4, when an *wildcard event listener* has been registered, the listener method will be called with two arguments:
- Event name
- Event payload

That said, following changes will be required in `SentryLaravelEventHandler::onWildcardEvent`.
1. Add previously mentioned arguments:
```php
public function onWildcardEvent($eventName, $payload)
```
2. Instead of `$args = func_get_args();` we can use the second argument which is the event payload. Eg.
```php
$route = $args[0]->route;
```
will have to be replaced with
```php
$route = $payload[0]->route;
```

While we're on the subject of the `onWildcardEvent` method it feels a bit messy for me. I refactored it [here](https://gist.github.com/KKSzymanowski/1c7965015e30c19e010fa3972db9b091), tell me what do you think.

Also, event names have changed. Eg. Laravel no longer raises `illuminate.log` event, instead an instance of `Illuminate\Log\Events\MessageLogged` is sent. You can see more in [this commit](https://github.com/laravel/framework/commit/57c82d095c356a0fe0f9381536afec768cdcc072). I reflected these changes in the Gist.

For now it's all I have discovered in this subject. Tell me if you're interested in a Pull Request regarding these changes.

stayallive:
It does not look like the handler in the Gist is backwards compatible with 5.3 and lower, is that also possible to implement somehow or would we need a different handler for <=5.3 and >=5.4?

KKSzymanowski:
Yes, it isn't compatible. I asuggest creating a separate branch for 5.4 support. I've seen couple of packages approach similar problems this way.

KKSzymanowski:
Although I just had an idea. If we create bespoke handlers for each event it should work on whole 5.*, because only wildcard API changed that much. I'm not sure though, I will check it later today.

PR #47, opened 2017-01-19
Prepare event handlers for Laravel 5.4

KKSzymanowski:
This pull request prepares `sentry-laravel` for the upcoming Laravel 5.4 release.

There are two changes that affect this package, as I described [here](https://github.com/getsentry/sentry-laravel/issues/46):
1. Different handling of *wildcard event* - the `Dispatcher::firing()` method has been removed and instead the event name is passed as the first argument.
2. The `illuminate.log` event has been changed to `Illuminate\Log\Events\MessageLogged`.

For the second change, I simply added another handler.
First change however was a bit more tricky. Because of changed event handling I was left with two options:
1. Detect Laravel version and act appropriately
2. Refactor *wildcard* event handlers to bespoke function for each event.

I chose the second one, simply because it provides cleaner and more easily extendable logic.
There are three events which need handling. For each of them there is an **old** and a **new** version, therefore there are total of 6 events with handlers attached:
```php
protected static $eventHandlerMap = [
    'router.matched'                           => 'routerMatched', // Until Laravel 5.1
    'Illuminate\Routing\Events\RouteMatched'   => 'routeMatched',  // Since Laravel 5.2

    'illuminate.query'                         => 'query',         // Until Laravel 5.1
    'Illuminate\Database\Events\QueryExecuted' => 'queryExecuted', // Since Laravel 5.2

    'illuminate.log'                           => 'log',           // Until Laravel 5.3
    'Illuminate\Log\Events\MessageLogged'      => 'messageLogged', // Since Laravel 5.4
];
```

Because wildcard events essentially went through one method, there was no problem with catching all exceptions thrown by handlers. Now however, as each event has it's own method there would be `try-catch` block duplication all over the class. In order to avoid that, I don't register actual handler in the Dispatcher, instead I register a fictional method, a call to which is intercepted by `__call` method, handling all exceptions and passing the event payload down to the actual handler. 

I have tested these changes in Laravel 5.1, 5.2, 5.3 and 5.4 with following bit of code:
```php
\Log::info('test');
\App\User::get();
echo($nonexistentVariable);
```
An `undefined variable` error was caugth by Sentry and in the Issue summary there was the correct route as well as log entry and SQL query.

5.0 doesn't support PHP 7 and isn't supported by Laravel team, so I didn't run any tests there. Also Laravel event logic hasn't been changed between 5.0 and 5.1 so all code that works with 5.1 should also work with 5.0.

### DIFF
--- a/src/Sentry/SentryLaravel/SentryLaravelEventHandler.php
+++ b/src/Sentry/SentryLaravel/SentryLaravelEventHandler.php
@@ -2,96 +2,191 @@
 
 namespace Sentry\SentryLaravel;
 
+use Exception;
+use Illuminate\Database\Events\QueryExecuted;
 use Illuminate\Events\Dispatcher;
+use Illuminate\Log\Events\MessageLogged;
+use Illuminate\Routing\Events\RouteMatched;
+use Illuminate\Routing\Route;
+use Raven_Client;
 
-// Event handling inspired by the ``laravel-debugbar`` project:
-//   https://github.com/barryvdh/laravel-debugbar
 class SentryLaravelEventHandler
 {
-    public function __construct(\Raven_Client $client, array $config)
+
+    /**
+     * Maps event handler function to event names.
+     *
+     * @var array
+     */
+    protected static $eventHandlerMap = [
+        'router.matched'                           => 'routerMatched', // Until Laravel 5.1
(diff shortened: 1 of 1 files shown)


dcramer:
At a glance this looks really good. I'll try to do a deeper review tomorrow and get this merged in.

Thanks!

PR #51, opened 2017-01-25
Force sentry to use the absolute git path

mrsimonbennett:
I was running into issues when running artisan from a directory that did not have a git repository. It was because sentry config `exec` was using the directory the command was executed in, not the path Laravel is installed. This simple fix forces it to use the correct path.


dcramer:
Thanks!

PR #54, opened 2017-02-08
Adds missing query bindings to breadcrumbs

BenConstable:
I've been using this package on a project, and I've noticed that the `breadcrumbs.sql_binding` config value currently isn't having an effect.

This is down to a couple of small issues - a missing `data` parameter in the `<= 5.1` handler, and an incorrect `isset()` check in the `>= 5.2` handler.

Let me know if you need any more info.


dcramer:
Thanks!

BenConstable:
No problem 👍

dcramer:
Pushed out 0.6.1 which includes these two fixes

BenConstable:
Great, thanks.

PR #55, opened 2017-02-21
Optional user_context

travoltron:
The current implementation assumes the default Laravel auth implementation. For anyone who uses auth schemes that aren't part of the core, this leaves lots of debris on the command line: `sentry.breadcrumbs error=Auth guard driver [api] is not defined.`

This change makes this optional behavior and doesn't alter it in any way.

### DIFF
--- a/src/Sentry/SentryLaravel/SentryLaravelServiceProvider.php
+++ b/src/Sentry/SentryLaravel/SentryLaravelServiceProvider.php
@@ -86,16 +86,18 @@ class SentryLaravelServiceProvider extends ServiceProvider
                 'app_path' => app_path(),
             ), $user_config));
 
-            // bind user context if available
-            try {
-                if ($app['auth']->check()) {
-                    $user = $app['auth']->user();
-                    $client->user_context(array(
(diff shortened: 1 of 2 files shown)


src/Sentry/SentryLaravel/SentryLaravelServiceProvider.php
```diff
@@ -86,16 +86,18 @@ public function register()
...
             ), $user_config));
 
-            // bind user context if available
-            try {
-                if ($app['auth']->check()) {
-                    $user = $app['auth']->user();
-                    $client->user_context(array(
-                        'id' => $user->getAuthIdentifier(),
-                    ));
+            if($user_config['breadcrumbs.user_context']) {
```
dcramer: we wouldnt want to call this breadcrumbs -- not sure if 'user_context' is descriptive enough.

maybe user_context.auto = True/False?

travoltron: That works for me. If you'd like I can write up a little blurb on this feature.

lintaba:
when will it be released?

PR #58, opened 2017-03-12
Added console logger exceptions

iamalirezaj:
Sentry console exception handler feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,47 @@ class SentryContext
 }
 ```
 
+## Adding to console
+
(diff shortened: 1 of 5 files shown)


stayallive:
The report function on your error handler is also called for artisan commands.

Just tested in the Laravel 5.4 example with a command that throws an exception nicely routes to Sentry.

stayallive:
Closing this since it is not needed 👍  Thanks anyway for taking the effort to make this PR!

PR #59, opened 2017-03-15
Converting to long form array syntax for poor SoBs using PHP 5.3

NoMan2000:
Yeah I know.  I know.  But we’re still stuck on a legacy server with legacy Laravel and need to get Sentry up and running on it.  Since most other place use the long-form array syntax, there were only a few small changes needed.

Signed-off-by: Michael Soileau <rsoileau@no-ip.com>


dcramer:
Thanks!

PR #62, opened 2017-04-04
Replace call to get user then id, with just id()

johnboy-leeds:
No need to retrieve the full user object to get the id from it as the auth guard interface `$app['auth']` has a method for this: https://github.com/laravel/framework/blob/5.4/src/Illuminate/Contracts/Auth/Guard.php#L29

### DIFF
--- a/src/Sentry/SentryLaravel/SentryLaravelServiceProvider.php
+++ b/src/Sentry/SentryLaravel/SentryLaravelServiceProvider.php
@@ -90,9 +90,8 @@ class SentryLaravelServiceProvider extends ServiceProvider
                 // bind user context if available
                 try {
                     if ($app['auth']->check()) {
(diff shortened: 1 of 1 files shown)


dcramer:
This unfortunately doesnt exist in all versions of laravel, so we'd want to feature detect this to avoid breaking existing behavior

stayallive:
This correct, this was implemented in Laravel 5.2 so might be better to leave it as is since the implementation in Laravel retrieves the full user object anyway.

> https://github.com/laravel/framework/blob/5.4/src/Illuminate/Auth/SessionGuard.php#L200

stayallive:
Since this nice Laravel "sugar syntax" is not available in every Laravel version we support and internally Laravel does the same thing as we currently do, I am going to close this.

PR #67, opened 2017-04-24
Add check if Sentry was loaded before reporting

stayallive:
Preventing extra exceptions when config has not been published for example (or old config is still cached).


dcramer:
Thanks!

Issue #34, opened 2016-10-24
laravel \Log::error('error') does not work


AidasK:
Hi, 
I am using default laravel installation and calling ` \Log::error('error')` does not send error message to the sentry. I can see this in the log file so it is working. If I throw an exception everything works as expected. Error is sent to the sentry and is logged to the file.

I am using sentry.io
laravel v5.2.43
sentry/sentry (1.5.0)
sentry/sentry-laravel (0.4.1)

AidasK:
Calling `app('sentry')->captureMessage('bbbb');` works too.

dcramer:
@AidasK We don't automatically capture logging events as issues in Sentry. You could certainly configure a Monolog handler for it, but it's often noisy and incorrect. I'd suggest creating a "logError" abstraction which does both captureMessage (ideally captureException if you can have an error object) and Log::error.

AidasK:
Ended up using this in EventServiceProvider:
```php
        \Log::listen(function ($level, $message, $context) {
            if ($level == 'debug') {
                $level = 'info';
            }
            $context['level'] = $level;
            if ($message instanceof \Exception) {
                app('sentry')->captureException($message, [], $context);
            } else {
                app('sentry')->captureMessage($message, [], $context);
            }
        });
```

dcramer:
I'm still not opposed to having some kind of functionality like this in core (optional, see https://github.com/getsentry/sentry-laravel/pull/38). We havent had time to prioritize building it ourselves, but maybe a quick solution is to simply document how to do it yourself.

sykezz:
In case anyone's still wondering about this, `Log` facade and `log()` helper works with Sentry when you configure [log channels](https://docs.sentry.io/platforms/php/laravel/#log-channels) on Laravel 5.6+. #38 won't be needed.

PR #69, opened 2017-05-24
Adjust README.me with how to setup Laravel Logging

sakalauskas:
I think it's worth putting this into README.md as It took me some time until I found how to set-up the logging.

Kudos to @AidasK in #34

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,7 @@ In order to capture Laravel Log messages and send these messages to Sentry you m
 namespace App\Providers;
 
 use Log;
-use App\Observers\UserObserver;
+use Exception;
 use Illuminate\Support\ServiceProvider;
 
 class AppServiceProvider extends ServiceProvider
@@ -211,22 +211,12 @@ class AppServiceProvider extends ServiceProvider
                 $level = 'info';
             }
             $context['level'] = $level;
-            if ($message instanceof \Exception) {
+            if ($message instanceof Exception) {
                 app('sentry')->captureException($message, [], $context);
             } else {
                 app('sentry')->captureMessage($message, [], $context);
             }
         });
-    }
-
-    /**
-     * Register the service provider.
-     *
-     * @return void
-     */
-    public function register()
-    {
-        //
     }
 }
 ```
--- a/README.md
+++ b/README.md
@@ -185,6 +185,61 @@ class SentryContext
 }
 ```
 
+## Send Logging messages to Sentry
+
+In order to capture Laravel Log messages and send these messages to Sentry you may register log listener in the boot method of one of your service providers. In this example, we'll register the observer in the AppServiceProvider.
+
+```php
+
+namespace App\Providers;
+
+use Log;
+use App\Observers\UserObserver;
+use Illuminate\Support\ServiceProvider;
+
+class AppServiceProvider extends ServiceProvider
+{
+    /**
+     * Bootstrap any application services.
+     *
+     * @return void
+     */
+    public function boot()
+    {
+        Log::listen(function ($level, $message, $context) {
+            if ($level == 'debug') {
+                $level = 'info';
+            }
+            $context['level'] = $level;
+            if ($message instanceof \Exception) {
+                app('sentry')->captureException($message, [], $context);
+            } else {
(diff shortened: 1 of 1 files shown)


AidasK:
```php

        \Log::listen(function ($level, $message, $context) {
            switch ($level) {
                case 'debug':
                    $level = 'info';
                    break;
                case 'critical':
                    $level = 'error';
            }
            $context['level'] = $level;
            if ($message instanceof \Exception) {
                app('sentry')->captureException($message, [], $context);
            } else {
                app('sentry')->captureMessage($message, [], $context);
            }
        });
```

This is my current code block. I haven't checked if all laravel and sentry error levels are mapped. 
These are all possible values in raven:
    const DEBUG = 'debug';
    const INFO = 'info';
    const WARN = 'warning';
    const WARNING = 'warning';
    const ERROR = 'error';
    const FATAL = 'fatal';
And these are in Laravel:
        'debug'     => MonologLogger::DEBUG,
        'info'      => MonologLogger::INFO,
        'notice'    => MonologLogger::NOTICE,
        'warning'   => MonologLogger::WARNING,
        'error'     => MonologLogger::ERROR,
        'critical'  => MonologLogger::CRITICAL,
        'alert'     => MonologLogger::ALERT,
        'emergency' => MonologLogger::EMERGENCY,

debug, critical,notice,alert,emergency should be mapped as well.

In the use case, I would include tags:
```php
\Log::error('bad response code', ['tags' => ['response_code' => 404]]);
```

sakalauskas:
@AidasK thanks, I mapped all of these now, although, the user should define these themselves e.g. to only send critical events to sentry.

stayallive:
Quoting @dcramer I don't think we are going to add this since Sentry is not a logging tool.

> We think of Sentry as an additional thing beyond logging due to the depth it goes into. People compare it to logging because its the concept they're familiar with. It's a "log", but not traditional logging.

https://github.com/getsentry/sentry-php/issues/404#issuecomment-285525468

And quoting https://sentry.io/vs/logging/:

> To summarize, Sentry works with your application logging infrastructure, often integrating directly. It does not replace the need for those logs, and it's also not a destination for things that aren't actionable errors or crashes.

AidasK:
```php
\Log::error('bad response code', ['tags' => ['response_code' => 404]]);
```
This looks like an error to me. Don't know why it should not be merged. I am paying for every log message and it is up to me what I am sending there.

stayallive:
You are absolutely free to do whatever you want, that was not what I meant to say. 

The reason why IMHO it should not be merged is because this explains how to do something that is not what Sentry wants you to do persé, use Sentry as a logging tool.

It might be better to only send warning/error (and worse) levels instead of sending every event level like the PR currently suggests.

This will keep it more in the direction of logging "actionable errors or crashes".

dcramer:
I'm not opposed to documenting this, but at the end of the day our goal is exception reporting. We *do* support capturing errors from things like logging (as "Message" events), I just try to continually enforce what the objective is so people dont conflate Sentry as being an alternative to something like ELK/Splunk/etc/etc

ankurk91:
This package may help in logging

https://github.com/jenssegers/laravel-raven

AidasK:
Laravel 5.5 has changed a lot.
You have to remove report method from `Handler` and add this to event service provider.
```php
        \Log::listen(function (MessageLogged $e) {
            $level = $e->level;
            switch ($level) {
                case 'debug':
                    $level = 'info';
                    break;
                case 'notice':
                case 'alert':
                    $level = 'warning';
                    break;
                case 'critical':
                    $level = 'error';
            }
            $context = $e->context;
            if (isset($context['exception'])) {
                $message = $context['exception'];
                unset($context['exception']);
            } else {
                $message = $e->message;
            }
            $context['level'] = $level;
            if ($message instanceof \Exception) {
                app('sentry')->captureException($message, [], $context);
            } else {
                app('sentry')->captureMessage($message, [], $context);
            }
        });
```

stayallive:
With the new Laravel 5.6 logging channels this should be implemented somewhat like you wanted possibly.

> https://github.com/getsentry/sentry-laravel#using-laravel-56-log-channels

Closing this for now.

PR #81, opened 2017-06-29
Fix auth context not working for Laravel >=5.3

### LINKED ISSUE #73: User context in laravel 5.4
Hi i would like to ask how you can add user context to sentry's entry when Auth::user() is null. This is related to laravel's auth handling for laravel versions > 5.2. Thanks

stayallive:
Because of changes in how the session is loaded in Laravel 5.3 we can no longer rely on getting the user inside the register method of the service provider. But there is an authenticated event introduced making it easy to simply listen for it and we can be sure the user is available.

Fixes #73

### DIFF
--- a/src/Sentry/SentryLaravel/SentryLaravelEventHandler.php
+++ b/src/Sentry/SentryLaravel/SentryLaravelEventHandler.php
@@ -3,12 +3,13 @@
 namespace Sentry\SentryLaravel;
 
 use Exception;
-use Illuminate\Database\Events\QueryExecuted;
+use Raven_Client;
+use Illuminate\Routing\Route;
 use Illuminate\Events\Dispatcher;
 use Illuminate\Log\Events\MessageLogged;
+use Illuminate\Auth\Events\Authenticated;
 use Illuminate\Routing\Events\RouteMatched;
-use Illuminate\Routing\Route;
-use Raven_Client;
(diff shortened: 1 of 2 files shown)


jconroy:
@stayallive @dcramer quick question - are these changes also needed for Lumen?

stayallive:
Hey @jconroy,

The `SentryLaravelEventHandler` we modified here is also used for Lumen (Lumen is basicly a stripped down version of Laravel) so the changes here also affect the Lumen integration.

I am sure you are asking because something is not working, could you open an issue with details if so? Than we can take a look why that is!

jconroy:
ok thanks @stayallive was just going off that the service providers are still different

```
src/Sentry/SentryLaravel/SentryLaravelServiceProvider.php
src/Sentry/SentryLaravel/SentryLumenServiceProvider.php
```

> I am sure you are asking because something is not working,

We just noticed we've been getting very odd (just straight incorrect) user id tagging in sentry and we can't put our finger on where the issue lies. We can see the events mention the correct user in the stack traces and breadcrumbs etc. but every so often the tag is wrong and it makes debugging very confusing. I'm not sure it could be to do with merging functionality in Sentry trying to be too smart or something else.

stayallive:
@jconroy this sounds like some events might occur before your own user integration can run and you see the default id set by Sentry, but please do make an new issue with what kind of errors this occurs so we can dig in.

PR #82, opened 2017-07-03
Fix syntax mistake in code example in readme.md

Spriz:


stayallive:
This is not a mistake, the `config/app.php` is not namespaced so the change you made is not required.

stayallive:
Hmm, not sure why that would be, maybe you are loading Service Providers or configuration in a custom way? It is not needed by default (see: https://github.com/laravel/laravel/blob/master/config/app.php#L138-L180).

PR #91, opened 2017-07-26
Use route URI for transaction when missing name

dcramer:


dcramer:
Aside I'm not 100% sure transaction route names are working in 5.1

dcramer:
updated, and confirmed it works in 5.1 (woo!)

PR #92, opened 2017-07-26
Basic CI integration

dcramer:


dcramer:
This doesn't do much, but it's a start.

test/Sentry/SentryLaravelServiceProviderTest.php
```diff
@@ -0,0 +1,31 @@
...
+use Sentry\SentryLaravel;
+
+
+class SentryLaravelServiceProviderTest extends \Orchestra\Testbench\TestCase
+
+{
+
+    protected function getEnvironmentSetUp($app)
+    {
+        // $app['config']->set('services.stripe.secret', getenv('STRIPE_SECRET'));
```
dcramer: left this as an example (from the blog post I pulled this from), though we could remove it

stayallive: Maybe we can set the sentry config with it and make sure the client takes on those configurations?

test/Sentry/SentryLaravelServiceProviderTest.php
```diff
@@ -0,0 +1,44 @@
...
+    protected function getPackageAliases($app)
+    {
+        return [
+            'Sentry' => 'Sentry\SentryLaravel\SentryFacade'
+        ];
+    }
+
+    public function testIsBound()
+    {
+        $this->assertEquals(app()->bound('sentry'), true);
```
dcramer: kind of silly since every test would fail if this one did

i imagine phpunit has dependencies but i didnt look

stayallive: In the phpunit.xml (and cli) you can configure it to stop running after the first failure, assuming this is the first test this is a way? (https://stackoverflow.com/a/30037442/1580028)

dcramer: found https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html

PR #96, opened 2017-08-11
Improve examples for prod usage

stayallive:
This improves the examples for production usage, showing the error page only when needed and converting non-http exception to a 500 error showing the `errors/500.blade.php` when `app.debug` is set to `false`. This makes it more production ready if code was taken from the examples directory.

Also added a section to the README.md to explain how you could show the error ID to users.

Reason: https://forum.sentry.io/t/laravel-auth-middleware/2107

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,30 @@ class SentryContext
 }
 ```
 
+## Displaying the error ID
+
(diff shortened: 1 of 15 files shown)


stayallive:
@dcramer looks like Laravel 4.2 used tabs instead of spaces, caused identation to weird out.

Should be fixed now :)

Issue #36, opened 2016-11-04
Issue running in production only on Laravel 4


samdking:
When I load in the service provider only on the production environment, by creating the following environment-specific config file in app/config/production/app.php:

```php
<?php

return [

	'providers' => [
		'Sentry\SentryLaravel\SentryLaravelServiceProvider',
	],

	'aliases' => [
		'Sentry' => 'Sentry\SentryLaravel\SentryFacade',
	],

];
```

Sentry breaks for logged in users at this point on line 85 in the service provider:

```php
                if ($app['auth']->check()) {
                    $user = $app['auth']->user();
                    $client->user_context(array(
                        'id' => $user->getAuthIdentifier(),
                    ));
                }
```

This is due to `SentryLaravelServiceProvider`'s boot method running first (by virtue of it being at the top of the list of service providers). Here is a stack trace received in public/error_log:

```
PHP Fatal error:  Call to a member function connection() on null in /path/to/app/bootstrap/compiled.php on line 7453
PHP Stack trace:
PHP   1. {main}() /path/to/app/public/index.php:0
PHP   2. Illuminate\Foundation\Application->run() /path/to/app/public/index.php:49
PHP   3. Stack\StackedHttpKernel->handle() /path/to/app/bootstrap/compiled.php:657
PHP   4. Illuminate\Cookie\Guard->handle() /path/to/app/bootstrap/compiled.php:11040
PHP   5. Illuminate\Cookie\Queue->handle() /path/to/app/bootstrap/compiled.php:8375
PHP   6. Illuminate\Session\Middleware->handle() /path/to/app/bootstrap/compiled.php:8428
PHP   7. Barryvdh\Debugbar\Middleware\Stack->handle() /path/to/app/bootstrap/compiled.php:7821
PHP   8. Illuminate\Foundation\Application->handle() /path/to/app/vendor/barryvdh/laravel-debugbar/src/Middleware/Stack.php:34
PHP   9. Illuminate\Foundation\Application->boot() /path/to/app/bootstrap/compiled.php:695
PHP  10. array_walk() /path/to/app/bootstrap/compiled.php:634
PHP  11. Illuminate\Foundation\Application->Illuminate\Foundation\{closure}() /path/to/app/bootstrap/compiled.php:634
PHP  12. Sentry\SentryLaravel\SentryLaravelServiceProvider->boot() /path/to/app/bootstrap/compiled.php:633
[...]

PR #99, opened 2017-08-11
Should use the ENV

lukepolo:
There are probably lots of cases where you want to track it in different places

### DIFF
--- a/src/config/config.php
+++ b/src/config/config.php
@@ -1,7 +1,7 @@
 <?php
 
 return array(
-    'dsn' => '',
+    'dsn' => env('SENTRY_DSN'),
 
     // capture release as git sha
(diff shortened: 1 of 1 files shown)


stayallive:
Actually we are already doing that. The file you changed is there for Laravel 4.2 compatibility.

This is the file that is published for `5.x` installations.

https://github.com/getsentry/sentry-laravel/blob/650526496b19982d7fe00f98efd02adea29ca4e4/src/Sentry/SentryLaravel/config.php#L3-L14

4.2 has a older discussion about adding some kind of environment config by default: #36

lukepolo:
ah , sorry thanks!

stayallive:
No problem at all, thank you for making the PR 😃

PR #107, opened 2017-10-08
Type hint interface instead of class

vojtasvoboda:
**Introduction:** I'm using OctoberCMS, which is made on the top of Laravel 5.5 framework.

**Problem:** OctoberCMS uses own implementation of Event Dispatcher and when using Sentry in OctoberCMS, I'm getting this error:

```
TypeError: Argument 1 passed to Sentry\SentryLaravel\SentryLaravelEventHandler::subscribe() must be an instance of Illuminate\Events\Dispatcher, instance of October\Rain\Events\Dispatcher given, called in /home/ubuntu/sites.com/vendor/sentry/sentry-laravel/src/Sentry/SentryLaravel/SentryLaravelServiceProvider.php on line 72
```

**Solution:** Type-hint interface instead of particular class.

### DIFF
--- a/src/Sentry/SentryLaravel/SentryLaravelEventHandler.php
+++ b/src/Sentry/SentryLaravel/SentryLaravelEventHandler.php
@@ -5,7 +5,7 @@ namespace Sentry\SentryLaravel;
 use Exception;
 use Raven_Client;
 use Illuminate\Routing\Route;
-use Illuminate\Events\Dispatcher;
+use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Log\Events\MessageLogged;
 use Illuminate\Auth\Events\Authenticated;
 use Illuminate\Routing\Events\RouteMatched;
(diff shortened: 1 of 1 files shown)


stayallive:
We would need to split our event handler to maintain 4.x support since 4.x does not implement those contracts (since they were introduced in 5.0). This change as is would break 4.x. 

Might be better to not typehint at all maybe?

vojtasvoboda:
How about creating special 4.x branch for a version compatible with old Laravel version. And cherry picking only bug fixes from master (L5.x) version?

stayallive:
Yeah, that would possibly be the better plan. Let me work that out.

stayallive:
Sorry for the delay. There will have to be some more cleanups before I release the `>5.x` version (need te remove all 4.x code), but you can use master or a specific commit for now to fix this particular issue.

vojtasvoboda:
Thanks! I'm using this:

```
"require": {
    "sentry/sentry-laravel": "dev-master#4166a9b"
}
```

and works well.

PR #109, opened 2017-11-09
Use existing class property instead of global helper in readme

kminek:


stayallive:
Is  there a specific reason for this? I like the other way since that code can be copied to anywhere without modifications and it can be easily changed to your format for the people that really hate the helper methods 😄

kminek:
it's a matter of taste really, one might argue that global helper functions are evil :)

stayallive:
I understand where you are coming from, thanks for taking the time to make the PR in the first place though!

PR #113, opened 2017-11-29
Only send exceptions to Sentry when debug is false

royduin:


stayallive:
I would rather advise to not supply a DSN at all in dev/local environments (where debug is enabled).

If the DSN is not set, Sentry won't try to send the exceptions that are captured.

Would you agree (ofcourse you are very free to implement this for yourself if needed but it should not be the recommended way)?

royduin:
But to debug locally if everything is coming through correctly I've supplied the DSN. So I could remove it after testing or add this so I can turn it on/off by changing `APP_DEBUG` in my `.env`. It's kind of personal preference.

stayallive:
I understand how this can be usefull in some workflows, but I don't think we should advertise it as "the way". Thanks for taking the time to make this PR and discuss the reasons for it!

PR #114, opened 2017-12-01
Also test against laravel 5.5

arubacao:
Travis now tests PHP versions `5.6`, `7.0`, `7.1`, `7.2` against all Laravel versions `>=5.0`.
Exception:

- Laravel `5.5` on PHP `5.6`
- Laravel `5.0` on PHP `7.2`
  
Results in 22 tests on Travis.

Cheers,
Chris

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,31 +1,38 @@
 language: php
+
 sudo: false
+
+cache:
+  directories:
+    - $HOME/.composer/cache
 php:
   - 5.6
   - 7.0
   - 7.1
   - 7.2
+
 env:
   matrix:
-  - LARAVEL=5.0 TESTBENCH=3.0.*
-  - LARAVEL=5.1 TESTBENCH=3.1.*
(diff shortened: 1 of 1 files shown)


stayallive:
Hey @arubacao, thanks for this. I have one question.

Why opt for removing `--prefer-dist`?

arubacao:
Hi @stayallive 
I removed `--prefer-dist` for no particular reason. I can add it back if you want to.
To be honest, I don't think it will make any difference because:

> For stable versions Composer will use the `dist` by **default**.

and

> `--prefer-dist`: Reverse of `--prefer-source`, Composer will install from dist if **possible**.
> This **can** speed up installs substantially on build servers and other use cases where you typically do not run updates of the vendors.

Seems to me if `--prefer-dist` is the default behavior anyway, but again, I'm not really sure^^.
Cheers,
Chris

stayallive:
Ah cool, well since it's the defautl and we won't require any non-stable packages it should be fine.

Thanks!

PR #121, opened 2018-02-08
Support Laravel 5.6

m1guelpf:
I did some initial tests, and the recommended setup seems to work out of the box.


stayallive:
I think you need to take another look at the changes you made 👍 

Laravel 5.6 was released... not 6.0 😉

Issue #119, opened 2018-02-07
Support for Laravel 5.6


straube:
Laravel 5.6 introduces the idea of [logging channels](https://laravel.com/docs/5.6/logging). It would be nice if Sentry implements that new feature.

antoniopaisfernandes:
There's even a problem with ```bindLogger``` implementation for 5.6 with the usage of ```getMonolog``` method that does not exist.

freekmurze:
Is the current version of the package able to send exceptions correctly to sentry (when installed in a L5.6 app)?

m1guelpf:
@freekmurze Yes

patrickbrouwers:
Submitted a PR for a sentry log channel: https://github.com/getsentry/sentry-laravel/pull/122

stayallive:
I'll close this since 5.6 is supported and PR #122 is dealing with adding the log channel 👍

PR #122, opened 2018-02-08
Add support for laravel 5.6 log channels

patrickbrouwers:
As #119 states Laravel 5.6 adds a nice new way of having log channels. 

Integration would be as easy as:

```
'sentry' => [
    'driver' => 'sentry',
    'level'  => 'notice',
],
```


antoniopaisfernandes:
Please replace ```$this->app::VERSION``` with ```$app::VERSION``` for Travis sake :)

dustingraham:
Laravel 5.6 requries PHP `^7.1.3` hence the travis failure.

stayallive:
Hey @patrickbrouwers, thanks for taking the time to create this PR 👍 sorry it took so long to get here :)

I am reading through the docs for custom log channels and seeing this:

> If you would like to define an entirely custom channel in which you have full control over Monolog's instantiation and configuration, you may specify a custom driver type in your  config/logging.php configuration file. Additionally, your configuration should include a via option which specifies the class that should be invoked to create the Monolog instance:

https://laravel.com/docs/5.6/logging#creating-custom-channels

Wouldn't it be nicer to simply have a `\Sentry\SentryLaravel\SentryLogChannel::class` that a user can add using the `custom` driver?

patrickbrouwers:
@stayallive that's also an option, but I'd personally prefer to enter a short string there. Also looking at how other packages handle this (https://docs.bugsnag.com/platforms/php/laravel/#reporting-unhandled-exceptions) it seems that's kind of the way to go.

stayallive:
@patrickbrouwers 👍 

Would you be willen to rename `CreateSentryLogChannel` to `SentryLogChannel`?

Apart from that LGTM! Thanks again!

patrickbrouwers:
@stayallive I can if you want, but it's not really a channel, but more of a factory. Let me know what you prefer.

stayallive:
Yes, you are totally right. However I think in the Laravel context this a how a channel is defined. So that's why `Create` sounds a bit weird to me. 

Also for people (for whatever reason) using the class directly in their config it would "look" better without the `Create` prefix IMHO.

So my vote would be to leave it out.

patrickbrouwers:
@stayallive , sure! I changed it.

patrickbrouwers:
@stayallive not sure why cs-fixer fails for php5.6 ?

stayallive:
You branch was a bit behind with older Travis / dependency thingies 😄

patrickbrouwers:
At least we have good old fashion isset checks still :)

stayallive:
Cool, that worked, will release this in the coming days (release for sentry-php base package also incoming, will release at the same time).

royduin:
Nice, but who's going to update the readme? It's still mentioning the "old" way...

stayallive:
@royduin what do you mean? It is documented...? Or is it missing things?

https://github.com/getsentry/sentry-laravel#using-laravel-56-log-channels

PR #123, opened 2018-02-12
Remove Laravel 4 support

stayallive:
Since Laravel 4 was released in 2013 and we have a good base for support (0.8.*) we can safely drop it IMO. So this PR does this and starts testing on Laravel 5.6.

phpcs fixer was also updated to v2 because of conflicts, I am not sure on the config but we can revisit this later and sync with `sentry-php`.


test/Sentry/SentryLaravelServiceProviderTest.php
```diff
@@ -20,7 +17,7 @@ protected function getPackageProviders($app)
     protected function getPackageAliases($app)
     {
         return [
-            'Sentry' => 'Sentry\SentryLaravel\SentryFacade'
+            'Sentry' => 'Sentry\SentryLaravel\SentryFacade',
```
Jean85: Maybe you can use `::class`?

composer.json
```diff
@@ -13,11 +13,11 @@
     ],
     "require": {
         "php": ">=5.3.0",
```
Jean85: Since this modification seems to break BC, maybe you should bump to `"php": "^5.6||^7.0"`? 
It's not ideal to have an open upper constraint...

.travis.yml
```diff
@@ -20,11 +20,16 @@ env:
   - LARAVEL=5.3.* TESTBENCH=3.3.* PHPUNIT=5.7.*
   - LARAVEL=5.4.* TESTBENCH=3.4.* PHPUNIT=5.7.*
   - LARAVEL=5.5.* TESTBENCH=3.5.* PHPUNIT=~6.0
+  - LARAVEL=5.6.* TESTBENCH=3.6.* PHPUNIT=~7.0
```
Jean85: Are you forced to specify the PHPUnit version too? It should be automatically chosen by Composer...

composer.json
stayallive: You are correct, good one!

However I have to make this `^5.4||^7.0` since Laravel 5 works on PHP 5.4.

.travis.yml
stayallive: Yes, since composer.json specifies `^4.6.6` and the other 2 dependencies also so they need to be overwritten in the composer file which is what these do.

test/Sentry/SentryLaravelServiceProviderTest.php
stayallive: No since we have to support PHP 5.4 :( (for Laravel 5.0)

dcramer:
Seems fine to me (dropping support), though I would ask we document "On Laravel 4? Use version XXX".

README.md
```diff
@@ -14,6 +14,10 @@
 
 Laravel integration for [Sentry](https://sentry.io/).
 
+## Laravel Version Compatibility
+
+- Laravel `4.2.x` is supported untill version `0.8.x` (`composer require "sentry/sentry-laravel:0.8.*"`)
```
Jean85: Typo: `until`

composer.json
```diff
@@ -13,14 +13,14 @@
     ],
     "require": {
         "php": "^5.4||^7.0",
-        "illuminate/support": "^5.0",
-        "sentry/sentry": "^1.8"
+        "illuminate/support": "5.0.*|5.1.*|5.2.*|5.3.*|5.4.*|5.5.*|5.6.*",
```
Jean85: Maybe `^5.0, <5.7` is more readable?

stayallive:
Fix issues with the slow failing builds thanks to @agentsib (getsentry/sentry-symfony#112)

src/Sentry/SentryLaravel/SentryLaravelServiceProvider.php
```diff
@@ -105,18 +85,16 @@ public function register()
...
-                        }
-                    } catch (\Exception $e) {
-                        error_log(sprintf('sentry.breadcrumbs error=%s', $e->getMessage()));
+                    if ($app['auth']->check()) {
+                        $client->user_context(array(
+                            'id' => $app['auth']->user()->getAuthIdentifier(),
+                        ));
                     }
+                } catch (Exception $e) {
+                    error_log(sprintf('sentry.breadcrumbs error=%s', $e->getMessage()));
```
mfn: Any chance this should be named `sentry.user_context` maybe, instead of …breadcrumbs?

PR #125, opened 2018-03-13
don't-discover value needs to be an array

Chris1904:
or else `in_array() expects parameter 2 to be array, string given`.

vendor/laravel/framework/src/Illuminate/Foundation/PackageManifest.php
line 117 - is where it’s affected.

This will simply fix the readme.


stayallive:
Thanks for this!

Issue #105, opened 2017-10-06
custom sentry config


drmax24:
We use sentry-laravel in our projects. However there is one problem.
We often call sentry like this:

`app(‘sentry’)->captureMessage(’bla-bla', [], ['level' => 'warning']);`

If called like this sentry reads configuration from file inside vendor dir.
We would like it to read our configuration file not yours.
/config/sentry.php would be ok

stayallive:
Hey @drmax24,

Have you followed the readme? It states you can publish the config file using:

```bash
php artisan vendor:publish --provider="Sentry\SentryLaravel\SentryLaravelServiceProvider"
```

This will create a config file in `config/sentry.php` and use that.

You can always copy the config file from the vendor to `config/sentry.php` yourself to achieve the same.

drmax24:
We have config/sentry.php but it i felt like 
app(‘sentry’)->captureMessage(’bla-bla', [], ['level' => 'warning']);
uses config inside vendior dir.

I'll check it again.

stayallive:
Well if the message is coming through, you are definitly using your custom config (since otherwise the DSN is not loaded and events are not being transmitted to Sentry).

stayallive:
Hey @drmax24 did you figure it out what was happening or are you still having difficulties?

stayallive:
Closing this because of inactivity, please do re-open if the problem persists!

drmax24:
I still have a problem.


Here is my config/sentry.php file:
<?php
```
return array(
    'dsn' => env('IS_SENTRY_ENABLED',false)?(env('SENTRY_DSN_BACK_END','')):'',

    // capture release as git sha
    // 'release' => trim(exec('git log --pretty="%h" -n1 HEAD')),

    // Capture bindings on SQL queries
    'breadcrumbs.sql_bindings' => true,
);
```

But sentry only reads SENTRY_DSN from .env

cjke:
This is a bug, perhaps, on your interpretation (imo it is - at minimum, it should be documented). It's due to the magic happening in the Raven_Client.php:

```
        if (!is_array($options_or_dsn) && !empty($options_or_dsn)) {
            $dsn = $options_or_dsn;
        } elseif (!empty($_SERVER['SENTRY_DSN'])) {
            $dsn = @$_SERVER['SENTRY_DSN'];
        } elseif (!empty($options['dsn'])) {
            $dsn = $options['dsn'];
        } else {
            $dsn = null;
        }
```

[...]

PR #130, opened 2018-04-26
Update SENTRY_DSN -> SENTRY_LARAVEL_DSN

stayallive:
This should fix the problems from #105.


m1guelpf:
@stayallive Isn't this a breaking change?

stayallive:
No. Since right now `SENTRY_DSN` in the .env always takes precedence whatever you do in the laravel config (see #105). 

So as long as you have the dsn set the old way that will apply untill you switch over if you need to. It is currently “broken” bit this change wont do anything for existing installations.

PR #132, opened 2018-04-30
fix: user_context default tracking to false

HazAT:
We decided to disable user tracking by default in all SDKs.
This is only to make sure users of the SDK really have to opt-in to track user information.


HazAT:
I've merged this now.
Knowing that this may be not the only thing we have to do.
We will need to add something to `sentry-php` too -> opt-in to ip address tracking.
But this still needs a definition because we want to streamline this setting in all SDKs.

PR #136, opened 2018-05-23
Update sentry DSN variable name in readme

mbaeuerle:


stayallive:
Hey @mbaeuerle, this is actually correct. The "default" env variable used to be `SENTRY_DSN` untill we notices it overwrites any variables set in `config/sentry.php` so we switched to `SENTRY_LARAVEL_DSN` so it would not overwrite the Laravel config 😄 Thanks for taking the effort to PR!

dcramer:
Is there a way to get back to SENTRY_DSN? This actually causes a problem with things like Heroku, as we bind SENTRY_DSN there and thats not something we can (or would want to) change.

stayallive:
@dcramer `SENTRY_DSN` still works just fine. Just anything in `config/sentry.php` is ignored when you set that `env` var which caused issues for some people.

Read from https://github.com/getsentry/sentry-laravel/issues/105#issuecomment-384162662 and down for the exact "issue".

So in short, setting `SENTRY_DSN` in your .env which is written to `$_ENV` is read by the PHP SDK and any other DSN config you pass to the client is thus ignored.

So this is very Laravel specific (or the `.env` implementation chosen by Laravel)

morrislaptop:
@stayallive I can confirm it can't find the DSN if only SENTRY_DSN is set in Laravel with the default config. 

The key for `dsn` is set to `null` and null is overriding the $_ENV variable.

PR #145, opened 2018-07-05
Pass logging config to monolog handler preparation to allow custom formatter

### LINKED ISSUE #142: Sentry Logging Time Issue
Since Laravel 5.6 there is an logging config file thus making it easier to get the application logs onto sentry. The issue I kept running into is that the log displays its time with the default formatter and each log with the same message creates a new issue in sentry.
I made a solution by creating a CustomFormatter class and tapped it into the sentry channel to change the formatter into one that truncates the time from the message.
Is there a cleaner and better way of solving this? Seems like a lot of ugly code for some basic functionality.
```php
<?php

namespace Certificates\Logging;

use Monolog\Formatter\LineFormatter;

class CustomFormatter
{
    /**
     * @param  \Illuminate\Log\Logger  $logger
     * @return void
     */
    public function __invoke($logger)
    {
        foreach ($logger->getHandlers() as $handler) {
            $handler->setFormatter(new LineFormatter("%channel%.%level_name%: %message% %context% %extra%\n",null, false,true));
        }
    }
}
```

```php
        'sentry' => [
            'driver' => 'sentry',
            'tap' => [Certificates\Logging\CustomFormatter::class],
        ],
```

Antipov-Anton:
Add possibility to set custom formatter in `config/logging.php` channels config.

Example of channel config:
```
        'sentry' => [
            'level' => 'error',
            'driver' => 'sentry',
            'formatter' => App\Logging\CustomFormatter::class,
        ],
```
Example of custom formatter:
```
<?php

namespace App\Logging;

use Monolog\Formatter\LineFormatter;

class CustomLineFormatter extends LineFormatter
{
    const SIMPLE_FORMAT = "%channel%.%level_name%: %message% %context% %extra%\n";
}
```
[Laravel documentation](https://laravel.com/docs/5.6/logging#creating-monolog-handler-channels) (See Monolog Formatters)

resolves #142

### DIFF
--- a/src/Sentry/SentryLaravel/SentryLogChannel.php
+++ b/src/Sentry/SentryLaravel/SentryLogChannel.php
@@ -21,6 +21,6 @@ class SentryLogChannel extends LogManager
             isset($config['bubble']) ? $config['bubble'] : true
         );
 
-        return new Logger($this->parseChannel($config), [$this->prepareHandler($handler)]);
+        return new Logger($this->parseChannel($config), [$this->prepareHandler($handler, $config)]);
     }
 }


stayallive:
Hey Anton, this is very simple, I like it! Thanks!

PR #146, opened 2018-07-17
Add tag with name of console command

ejunker:
Listen for the `CommandStarting` event and add a tag with the name of the console command. This makes it easy to know which console command has the error.

Note: I did not typehint `ComandStarting` in `commandStartingHandler` because then I think it would only work on Laravel >=5.5 and based on the documentation this package supports Laravel 5.*. I'm surprised that the typehint of `MessageLogged` on  `messageLoggedHandler` does not cause issues for people trying to use older versions of Laravel since `MessageLogged` was not added until Laravel 5.4.


stayallive:
Hey @ejunker, thanks for taking the time to make this PR.

Because of autoloading you can `use` a non-existing class and that won't cause problems since the method is never called causing the class to resolve and not be found. So I added the type-hint cause why not :)

Also I added the CommandFinished event to reset the command in case it was called from a web request, sounds good?

PR #152, opened 2018-09-25
fix readme with name attribute for logging channels in laravel 5.6+

smknstd:


stayallive:
Thanks @smknstd, I hope you won't mind I updated it a bit to fix a spelling error and to rewrite it so it doesn't remove the initial setup example.

I also removed the screenshot since Sentry changes so much this going to be outdated in a second. I think it's clear like this too.

Please let me know if you see issues still, otherwise this can be merged :)

smknstd:
@stayallive :+1:

PR #166, opened 2018-11-26
feat: Use laravel client and options

HazAT:
Lumen and Laravel need to be `>= 5.6` because of our requirement of PHP `>= 7.1`
It also currently uses a local overwrites in the PHP SDK to make it work.


Jean85:
> Lumen and Laravel need to be >= 5.6

Are you sure about that? Couldn't you have older versions of Laravel under PHP 7.1 and be happy about that?

PR #176, opened 2019-01-16
Hide error message with enabled release option

bodva:
In the different environments, git folder can exist or not.
If this part uncommented but git folder does not exist, this command throws a message in the console with error:
`fatal: Not a git repository (or any parent up to mount point /var/www)`
My changes hide this error even if the .git folder does not exist.

### DIFF
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -4,7 +4,7 @@ return array(
     'dsn' => env('SENTRY_LARAVEL_DSN'),
 
     // capture release as git sha
-    // 'release' => trim(exec('git --git-dir ' . base_path('.git') . ' log --pretty="%h" -n1 HEAD')),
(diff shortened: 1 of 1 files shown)


stayallive:
I don't think we should do this, the release should be set to an output without errors, if that errors for you you can replace it with a valid command or a static version number.

bodva:
@stayallive then you should show what the source of the error, because i spent a few hours to find where the hell my app call the git

bodva:
@stayallive  I am not sure if this is the place where you need to see the error about executing this command, but I understand your concern.

Maybe you need to try to catch the error in the place when you execute `release` command and add additional information about executing problem.

I don't think this additional comment can help with that, because of this obvious requirement.

PR #177, opened 2019-01-16
Hide error message with enabled release option

bodva:
In the different environments, git folder can exist or not.
If this part uncommented but git folder does not exist, this command throws a message in the console with error:
`fatal: Not a git repository (or any parent up to mount point /var/www)`
My changes hide this error even if the .git folder does not exist.


stayallive:
Closing this in favor of the one targetted at master (#176)

PR #179, opened 2019-01-18
Added missing queueJobProcessedHandler method

### LINKED ISSUE #178: PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 40960 bytes)
Greetings,

I am having a problem with sentry laravel when I start my queue with  `php artisan queue:work` I get the following error.

I am using the default config, and have set up the DSN in the .env file

```
D:\karugi>php artisan queue:work
[2019-01-16 18:34:13][554] Processing: App\Notifications\AdminNewAssessmentRequestedNotification
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 40960 bytes) in D:\karugi\vendor\sentry\sentry-laravel\src\Sentry\Laravel\EventHandler.php on line 102
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 32768 bytes) in D:\karugi\vendor\symfony\debug\Exception\FatalErrorException.php on line 1
```
Seems like this is the culprit inSentry\Laravel\EventHandler from line 99
```
/**
 * Pass through the event and capture any errors.
 *
 * @param string $method
 * @param array  $arguments
 */
public function __call($method, $arguments)
{
    try {
--->        call_user_func_array(array($this, $method . 'handler'), $arguments);
    } catch (Exception $exception) {
        // Ignore
    }
}
```
Don't really have a clue what's happening... If you need more info

Humni:
I'm not sure why this method was missing or if it was intentional, but adding the method in will resolve #178. The only thing modified in the method compared to the `queueJobProcessingHandler` is the 4th parameter in the breadcrumb creation. Could easily be refactored to reduce the code duplication.

### DIFF
--- a/src/Sentry/Laravel/EventHandler.php
+++ b/src/Sentry/Laravel/EventHandler.php
@@ -224,6 +224,34 @@ class EventHandler
     }
 
     /**
+     * Since Laravel 5.2
+     *
+     * @param \Illuminate\Queue\Events\JobProcessed $event
+     */
+    protected function queueJobProcessedHandler(JobProcessed $event)
+    {
+        $job = [
+            'job' => $event->job->getName(),
+            'queue' => $event->job->getQueue(),
+            'attempts' => $event->job->attempts(),
+            'connection' => $event->connectionName,
+        ];
+
+        // Resolve name exists only from Laravel 5.3+
+        if (method_exists($event->job, 'resolveName')) {
+            $job['resolved'] = $event->job->resolveName();
+        }
+
+        Integration::addBreadcrumb(new Breadcrumb(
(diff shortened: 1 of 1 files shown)


stayallive:
Sorry that is not the way it's meant to be.

If you look at this: https://github.com/getsentry/sentry-laravel/commit/6fdeaa068574c6202b44692181bb52e74b4bfc5e

It's supposed to clear any queued errors and breadcrumbs from the SDK to the next job starts clean again.

Humni:
Cool, not worried just need it fixed. Thanks for looking into it!

PR #187, opened 2019-02-07
Re-add more missing handlers

### LINKED ISSUE #186: "Missing event handler:loghandler" on Laravel 5.2 clean installation.
This is happening on Laravel 5.2, clean install.

`RuntimeException: Missing event handler:loghandler ... on line 104`

I was able to work around locally by adding `return;` right before the exception (src/Sentry/Laravel/EventHandler.php:104).

I saw that the mentioned exception is added on beta3. Downgrading to beta2 produces a memory exhaust right there, I'm guessing thats what the exception is for.

stayallive:
Something went wrong when the EventHandler was migrated I think...

Added 2 more missing handlers... should be complete now!

Fixes #186

### DIFF
--- a/src/Sentry/Laravel/EventHandler.php
+++ b/src/Sentry/Laravel/EventHandler.php
@@ -165,10 +165,12 @@ class EventHandler
             $data['bindings'] = $bindings;
         }
 
-        $this->record(array(
-            'message' => $query,
-            'category' => 'sql.query',
-            'data' => $data,
+        Integration::addBreadcrumb(new Breadcrumb(
+            Breadcrumb::LEVEL_INFO,
+            Breadcrumb::TYPE_USER,
+            'sql.query',
+            $query,
+            $data
         ));
     }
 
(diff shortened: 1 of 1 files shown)


src/Sentry/Laravel/EventHandler.php
```diff
@@ -171,6 +194,23 @@ protected function queryExecutedHandler(QueryExecuted $query)
...
+    /**
+     * Until Laravel 5.3
+     *
+     * @param string     $level
+     * @param string     $message
+     * @param array|null $context
+     */
+    protected function logHandler($level, $message, $context)
+    {
+        $this->client->breadcrumbs->record(array(
```
HazAT: Sorry, misclicked the approve.
We do not have `$this->client` anymore instead you need to call 
```
Integration::addBreadcrumb(
```

src/Sentry/Laravel/EventHandler.php
```diff
@@ -149,6 +149,29 @@ protected function routeMatchedHandler(RouteMatched $match)
...
+     */
+    protected function queryHandler($query, $bindings, $time, $connectionName)
+    {
+        $data = array('connectionName' => $connectionName);
+
+        if ($this->sqlBindings) {
+            $data['bindings'] = $bindings;
+        }
+
+        $this->record(array(
```
HazAT: There is no `$this->record`
You need to add breacrumbs like `Integration::addBreadcrumb(`

stayallive:
@HazAT sorry about that, I just copied them back without thinking more than 2 seconds. They should be good now. Even did a bit of testing on 5.2 for it 👍

PR #189, opened 2019-02-14
Do not lazy load Sentry on resolving

stayallive:
This way the Hub has the current (configured) client and users can use the Hub or SDK methods to configure their environment and/or capture exception with the already configured client.

@HazAT There is 1 discussion item about this. The user context (ID) is not captured automatically anymore unless you are on Laravel 5.3+ (there is an `authenticated` event we are listening for) since we cannot know when the session has initialized and it certainly is not in the register method of a service provider. Thought on trying to find a workaround or taking it as a breaking change and adding documentation on how to add the correct user context yourself!?

Not 100% convinced about moving the initialization to the boot method (context: https://laracasts.com/discuss/channels/general-discussion/difference-between-boot-and-register-method#reply=4507) but otherwise the test cannot work either so I think it makes sense.


HazAT:
@stayallive I think it's fine if we document that users need to set <5.3 need to set the user into the scope themselves.

PR #195, opened 2019-02-22
Update to allow Laravel 5.8

squiaios:
Update illuminate/support & laravel/framework to 5.8

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,12 +13,12 @@
   ],
   "require": {
     "php": "^7.1",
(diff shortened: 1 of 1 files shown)


stayallive:
I've removed `5.8` from the require-dev because it makes no sense (since that version doesn't even exist yet) so we are sure which version is installed in dev.

Looked into adding it to the test matrix but that's not possible until a new release of orchestra testbench is tagged.

From manual testing and reading the changelog so far I think it's safe to allow it 👍

PR #200, opened 2019-02-27
Support for Laravel 5.8

damcclean:
I've updated some of the dependencies in this package so that the Sentry package for Laravel supports the latest version, 5.8 which was released yesterday.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,9 +17,9 @@
     "sentry/sdk": "^2.0"
   },
   "require-dev": {
-    "phpunit/phpunit": "7.3.*",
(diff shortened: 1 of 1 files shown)


stayallive:
Thank you for taking the time but Laravel 5.8 is already supported on `1.0.0` 👍 

These are just the development dependencies that you updated which will not be installed on the end-user side and thus have no effect on the "compatibility". These will be updated later if we also update the test suite to include 5.8 in Travis CI.

Currently though only version `1.0.0` supports Laravel 5.8, `0.11.0` or lower does not.

PR #205, opened 2019-03-01
Prevent registering events when no DSN is set

stayallive:
This is #157 but for 1.0.

### DIFF
--- a/src/Sentry/Laravel/ServiceProvider.php
+++ b/src/Sentry/Laravel/ServiceProvider.php
@@ -25,7 +25,9 @@ class ServiceProvider extends IlluminateServiceProvider
     {
         $this->app->make(self::$abstract);
 
-        $this->bindEvents($this->app);
+        if ($this->hasDsnSet()) {
+            $this->bindEvents($this->app);
+        }
 
         if ($this->app->runningInConsole()) {
             if ($this->app instanceof Laravel) {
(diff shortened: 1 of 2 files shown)


src/Sentry/Laravel/ServiceProvider.php
```diff
@@ -117,6 +119,16 @@ protected function configureAndRegisterClient(): void
         });
     }
 
+    /**
+     * Check if a DSN was set in the config.
+     *
+     * @return bool
+     */
+    protected function hasDSNSet(): bool
```
HazAT:
```suggestion
    protected function hasDsnSet(): bool
```

src/Sentry/Laravel/ServiceProvider.php
```diff
@@ -25,7 +25,9 @@ public function boot(): void
     {
         $this->app->make(self::$abstract);
 
-        $this->bindEvents($this->app);
+        if ($this->hasDSNSet()) {
```
HazAT:
```suggestion
        if ($this->hasDsnSet()) {
```

test/Sentry/ServiceProviderWithoutDSNTest.php
```diff
@@ -0,0 +1,40 @@
...
+
+    public function testIsBound()
+    {
+        $this->assertTrue(app()->bound('sentry'));
+    }
+
+    /**
+     * @depends testIsBound
+     */
+    public function testDSN()
```
HazAT:
```suggestion
    public function testDsn()
```

PR #207, opened 2019-03-08
Fixing config syntax

ChrisThompsonTLDR:
This wasn't following Laravel syntax making it impossible to overload.

This is the normal way to overload a config param in Laravel
```
config(['sentry.breadcrumbs.sql_bindings' => false]);
```
this uses dot syntax.

The problem is that the config/sentry.php file uses `breadcrumbs.sql_bindings` which won't be overwrite-able with Laravel.

```
dump(config('sentry'));
config(['sentry.breadcrumbs.sql_bindings' => false]);
dd(config('sentry'));
```
will generate
```
array:2 [
  "dsn" => ""
  "breadcrumbs.sql_bindings" => true
]
array:2 [
  "dsn" => null
  "breadcrumbs.sql_bindings" => true
  "breadcrumbs" => array:1 [
    "sql_bindings" => false
  ]
]
```
The config param wasn't overwritten because `config()` can't accept a config param with a `.` in the name, it will map that to a multi-dimensional array.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ jobs:
 
 before_install:
   - if [ "$USE_COMPOSER_JSON" != "1" ]; then composer remove friendsofphp/php-cs-fixer --dev --no-update; fi;
-  - if [ "$USE_COMPOSER_JSON" != "1" ]; then composer require laravel/framework:$LARAVEL illuminate/support:$LARAVEL orchestra/testbench:$TESTBENCH phpunit/phpunit:$PHPUNIT php-http/curl-client guzzlehttp/psr7 --no-update --no-interaction --dev; fi;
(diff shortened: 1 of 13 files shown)


stayallive:
I would say yes and no to this change for the following reasons:

- Yes, you are correct, with that `.` in the name it's impossible to set using the `config` classes provided by Laravel.

- No, because it has no use to change from user code... after your application has loaded this configuration has been read and changing it will not effect the registered events.

You can ofcourse still override it from your own `config/sentry.php` config file (you can publish with `php artisan vendor:publish --provider="Sentry\Laravel\ServiceProvider"`).

ChrisThompsonTLDR:
@stayallive I'd be interested in having you do the tests.  I'm not super familiar with writing tests and would appreciate seeing them written by someone that knows what they are doing.

I wasn't sure how to make it backwards compatible other than checking two different keys (old key with dot syntax and new key with array).

stayallive:
So I won't claim I'm an expert on Laravel package testing but I think I came up with a good way to test this. Let me know what you think.

PR #217, opened 2019-03-21
Clarify which version supports which Laravel / PHP combination

baijunyao:


stayallive:
Hey @baijunyao,

We still support all Laravel 5 versions (as long as they can run on PHP 7.1 and higher) in the most recent version and we support `5.7` and below in version `0.11.x` for PHP versions below 7.1. 

We never stopped supporting 5.6 & 5.7.

Would you be willing to update your change to reflect that? Possibly something like:

```
- Laravel `5.7.x` and below on PHP 7.0 and below is supported in version `0.11.x`
- Laravel `5.x.x` on PHP 7.1 and above is supported in the most recent version
```

baijunyao:
Sorry, I modified.

PR #221, opened 2019-03-27
Updated provides.json

miladkiaei:
fixed namespaces for providers and aliases

### DIFF
--- a/provides.json
+++ b/provides.json
@@ -1,11 +1,11 @@
 {
   "providers": [
-    "Sentry\SentryLaravel\SentryLaravelServiceProvider"
+    "Sentry\Laravel\ServiceProvider"
   ],
   "aliases": [
     {
       "alias": "Sentry",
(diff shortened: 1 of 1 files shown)


stayallive:
@miladkiaei on what Laravel version are you?

Laravel 5.5 reads those values from the `composer.json` file and those are correct and Sentry correctly registers with Laravel... are you using another installer/detecter which uses the provides.json?

miladkiaei:
I'm using laravel 5.8. After updating framework and related packages, I got this error on package discover command.

stayallive:
@miladkiaei can you show me the error and your composer.json (the require/require-dev section is enough) so we can see if we can reproduce? 

I am not getting this error on multiple projects ranging from 5.5 to 5.8.

stayallive:
@miladkiaei I did some digging and it seems this is `provides.json` was for a Laravel 4 as far as I can tell. So we are removing this file altogether since Laravel reads the `composer.json` for the service provider & facades.

PR #224, opened 2019-04-03
Fallback to SENTRY_DSN if defined in env

morrislaptop:
#136 #105 #130

### DIFF
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -1,4 +1,7 @@
 <?php
+
+return [
+
     'dsn' => env('SENTRY_LARAVEL_DSN', env('SENTRY_DSN')),
 
     // capture release as git sha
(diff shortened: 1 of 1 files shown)


stayallive:
Can you show me where we still reference SENTRY_DSN?

This was changed a long time ago to use SENTRY_LARAVEL_DSN. 

https://docs.sentry.io/platforms/php/laravel/#laravel-5x

morrislaptop:
SENTRY_DSN is set as the default option in https://github.com/getsentry/sentry-php/blob/master/src/Options.php#L691

But if SENTRY_LARAVEL_DSN is not in the env, then we are essentially passing in:

```
$options = new Sentry\Options(['dsn' => null]);
```

The null overrides the default. 

I would expect that `$options->getDsn() === env('SENTRY_DSN')`

But, `$options->getDsn() === null`

PR #226, opened 2019-04-08
📄 Add phpdoc methods to the facade

erikgaal:
This PR adds `@method` doclets to the Facade class. This allows IDE's like PHPStorm to typehint usages of the `\Sentry\Laravel\Facade` class.


HazAT:
Thx!

PR #227, opened 2019-04-11
chore: consistent array syntax

chuangbo:


stayallive:
@chuangbo there are a lot of missed `array()` syntaxes left, not just in this file.

Would be nice to clean them up all at once if possible!?

chuangbo:
@stayallive sure will do

PR #228, opened 2019-04-11
Fix events generated in queue worker not sending

### LINKED ISSUE #222: Errors from queued jobs aren't sent until the queue runner is restarted
Exceptions from queued background jobs are logged properly, but they only seem to be sent to the remote Sentry server when you manually restart or shut down the queue runner. I assume that's the intended behavior for HTTP requests, but for queue runners it means reports are delayed indefinitely (or lost).

I'm using the Horizon queue runner, but I assume this happens with the basic one as well.

I think this was handled in #153, but the `sendUnsentErrors` method was removed from a later version of sentry-php so I'm not sure what the best method to fix this would be at this point.

stayallive:
This is a workaround until we are able to restructure / fix the underlying issue in the client and our inability to wait for all events to be sent (or make sure they are sent immediately instead of waiting for a `wait` call, context getsentry/sentry-php#799).

The scope popping was also not happening correctly because the `JobProcessed` event is only fired when a job is successfully handled. Another event is dispatched when an error occurs but the exception is logged after those events fire, thus the scope will already be popped once the exception is recorded, making the scope effectively useless.

/fixes #222


src/Sentry/Laravel/Integration.php
```diff
@@ -82,4 +84,30 @@ public static function setTransaction($transaction): void
...
+     * Block until all async events are processed for the HTTP transport.
+     *
+     * @internal This is not part of the public API and is here temporarily until
+     *  the underlying issue can be resolved, this method will be removed.
+     */
+    public static function flushEvents(): void
+    {
+        $client = Hub::getCurrent()->getClient();
+
+        if ($client instanceof Client) {
```
HazAT: We should implement `close`/`flush` in the main SDK, this should make this obsolete.
This code should go directly in there.

stayallive: @HazAT yes, we had extensive discussion about it and the consensus is we made a mistake implementing the async stuff, so this is a workaround until we can fix that in the base SDK 👍 this will be removed at a later time since it won't be needed if fixed in the base SDK.

PR #231, opened 2019-04-18
Switch the order of null coalescing for the sql_bindings config option

ObliviousHarmony:
Hello,

A breaking change was introduced in https://github.com/getsentry/sentry-laravel/pull/207.

The user and package config files are merged by the service provider, and so the order of the null coalescing means that the default of "true" is used, regardless of what the user set the legacy config setting to.


stayallive:
Thanks for catching this!

PR #239, opened 2019-05-13
Resolve integrations from container

### LINKED ISSUE #238: How to register a new integration outside config file?
As far as I know, there are two ways to register a new integration into the client:

1. Adding integration instances to `integrations` array in `config/sentry.php`;
2. Override `\Sentry\Laravel\ServiceProvider::configureAndRegisterClient` method and add integration instances at this line https://github.com/getsentry/sentry-laravel/blob/master/src/Sentry/Laravel/ServiceProvider.php#L113 .

But both ways are not good enough:

1. The custom integration may depend on other third-party objects which are not able to instantiate in the config file
2. When I copy-paste the body of `\Sentry\Laravel\ServiceProvider::configureAndRegisterClient` method into my own ServiceProvider class, it told me that `ClientBuilder::setSdkIdentifier` and `ClientBuilder::setSdkVersion` method are internal (they are not defined as `private` but there is `@internal` annotation).

So I wonder if there is an elegant way to add a new integration which has third-part dependence?

stayallive:
This will resolve the `integrations` option with the Laravel container (if needed).

I do not check if the instance retrieved from the container is an object that implements `IntegrationInterface` since the Sentry client will validate that itself.

This also makes the `Sentry\Laravel\Integration` not overwritable by the user which seems like a good side effect to me since there should be no reason to not register that integration.

/fixes #238

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Resolve `integrations` option from the container (#239)
+
 ## 1.0.2
 
 - Track Artisan command invocation in breadcrumb (#232)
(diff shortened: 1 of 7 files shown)


leo108:
Really need this, when can this be reviewed and merged?

PR #258, opened 2019-07-30
Add more options to breadcrumbs configuration

### LINKED ISSUE #257: Options to disable SQL and log breadcrumb
We don't want to send SQL and logs to Sentry, it there a way to disable them?

leo108:
Fixes #257

### DIFF
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -8,10 +8,17 @@ return [
     // 'release' => trim(exec('git --git-dir ' . base_path('.git') . ' log --pretty="%h" -n1 HEAD')),
 
     'breadcrumbs' => [
+        // Capture Laravel logs in breadcrumbs
+        'logs' => true,
+
+        // Capture SQL queries in breadcrumbs
+        'sql_queries' => true,
 
         // Capture bindings on SQL queries logged in breadcrumbs
         'sql_bindings' => true,
 
+        // Capture queue job information in breadcrumbs
+        'queue_info' => true,
     ],
 
 ];
--- a/src/Sentry/Laravel/EventHandler.php
+++ b/src/Sentry/Laravel/EventHandler.php
@@ -64,11 +64,32 @@ class EventHandler
     private $events;
 
     /**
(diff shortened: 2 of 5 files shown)


leo108:
@stayallive any feedback on the PR?

leo108:
@stayallive Is there any chance to look at this PR?

stayallive:
Yeah I'm really sorry for ignoring you for this long.

I don't think this is the way to go, since this does not allow us (or Laravel) to ever make changes in the event names/types we want to watch for an operation.

There are also quite some other cases where right now it's needed that the `EventHandler` is registered no matter what since otherwise events when running a queue worker might never be sent.

How about we do something like you requested in your original request. Just allow for disabling the query breadcrumbs?

We already have:

```php
    'breadcrumbs' => [
        // Capture bindings on SQL queries logged in breadcrumbs
        'sql_bindings' => true,
    ],
```

What about adding an option like for example:

```php
    'breadcrumbs' => [
        // Log SQL queries as breadcrumbs
        'queries' => true,

        // Capture bindings on SQL queries logged in breadcrumbs
        'sql_bindings' => true,
    ],
```

This way we could keep adding options to disable built-in breadcrumb features without having to have a defined list of events in the user config.

leo108:
@stayallive Thanks for the feedback, I'll check and update my PR.

stayallive:
Many thanks for taking the time to do that!

leo108:
@stayallive Please check again, thanks.

leo108:
@HazAT Can you check this PR? thanks.

nauxliu:
I would love to see this merged. How can we move forward on this PR?

stayallive:
@nauxliu: @HazAT was just pinged a few short hours ago to leave his review, after he is done and there are no more "problems" found it should al go pretty quick.

It's being worked on!

Thank you (both) for your patience.

nauxliu:
Great! looks we are making a progress, Thank you all @leo108 @stayallive @HazAT

PR #267, opened 2019-08-30
Adds support for Laravel ^6.0

plunkettscott:
Laravel 6.0 is released on September 3rd. This adds support for that release. I'm not sure how you want to handle versioning. There's normally different releases for the various versions, however, version 6 does not have any breaking changes that will affect Sentry's reporting, only that versioning is changing to semantic versioning starting at 6.0.x.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
   },
   "require-dev": {
     "phpunit/phpunit": "7.5.*",
-    "laravel/framework": "5.8.*",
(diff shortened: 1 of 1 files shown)


stayallive:
Thanks for taking the time, but this is not the correct way to go about this (you only added it to require-dev which is for development only and has no effect on if users can install it on that version).

Since we probably will get a lot of PR's I am going to close this (there is already another one) adding the needed changes to support 6 by the 3rd. We're on it!

PR #269, opened 2019-08-31
Add support for Laravel 6

erikgaal:
Creating this PR in advance of the Laravel 6 release. This PR add an entry to the travis test matrix for the latest release. As of `orchestra/testbench@3.9.x`, they require PHPUnit 8. This brings some _inconveniences_ with supporting older versions of PHPUnit in the same codebase.

- [x] PHPUnit 8 made `@expectedException` annotations deprecated. However, PHPUnit 4 doesn't have the `$this->expectException()` method yet, which is still used for testing with Laravel 5.0. Should we remove Laravel 5.0 form the test matrix, or drop support?
- [x] PHPUnit 8 and Laravel 6 require PHP >=7.2, which results in one more exception in the test matrix. (fixed in ce244f9)
- [x] There also was one breaking change (https://github.com/laravel/framework/pull/27977) that results in a test failing. (fixed in 0254084)
- [ ] Remove the `minimum-stability` and `prefer-stable` flags from the `composer.json` before merging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,12 +17,15 @@ env:
     - LARAVEL=5.6.* TESTBENCH=3.6.* PHPUNIT=7.5.*
     - LARAVEL=5.7.* TESTBENCH=3.7.* PHPUNIT=7.5.*
     - LARAVEL=5.8.* TESTBENCH=3.8.* PHPUNIT=7.5.*
+    - LARAVEL=6.0.* TESTBENCH=3.9.* PHPUNIT=8.0.*
 
 matrix:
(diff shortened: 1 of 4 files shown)


stayallive:
RE `@expectedException`: We'll leave it for now, there are just warnings, removal is happening in PHPUnit 9.

I do want to wait a bit to see if testbench is released before we cut ours, but if needed we'll release without having it in the test matrix for now (since local testing turned up no problems).

erikgaal:
@stayallive I've reverted the annotations, tests are now green.

PR #271, opened 2019-09-06
Flush events when Application terminates

ryzr:
Environment:
- Laravel 6.0
- PHP 7.3
- Using stack driver with Sentry and Single handlers

Exceptions thrown during artisan commands have not been reaching Sentry. Even `php artisan sentry:test` has been failing for me. Upon investigating, it looks like `exit($status)` in artisan prevents the `__destruct()` magic method in HttpTransport from cleaning up pending requests? Could be wrong here, but commenting out exit() did seem to fix the issue.

https://github.com/laravel/laravel/blob/master/artisan#L53
https://github.com/getsentry/sentry-php/blob/master/src/Transport/HttpTransport.php#L66

My proposed change is to hook into the Laravel application `terminating()` callback. This utilises the fix introduced here: https://github.com/getsentry/sentry-laravel/pull/228 to ensure all pending promises resolve.

Unfortunately, it doesn't look like Lumen offers equivalent functionality.

### DIFF
--- a/src/Sentry/Laravel/ServiceProvider.php
+++ b/src/Sentry/Laravel/ServiceProvider.php
@@ -60,6 +60,12 @@ class ServiceProvider extends IlluminateServiceProvider
                 return (new LogChannel($app))($config);
             });
         }
+
+        if ($this->app instanceof Laravel) {
+            $this->app->terminating(function () {
+                Integration::flushEvents();
+            });
+        }
     }
 
     /**


stayallive:
@ryzr thanks for taking the time to create this PR and investigate!

We have come to the same conclusion as you and it looks like a PHP bug causes this behaviour, since this affects not only Laravel we want to fix this upstream in getsentry/sentry-php#877 so we have a fix in place for every instance the PHP SDK is used.

Once this is tagged a composer update should be enough to fix it without any changes to sentry-laravel.

ryzr:
Brilliant, thanks for the response :-)

tlaverdure:
Has this been released? I'm not getting an errors with Laravel 6 running on Laravel Vapor

stayallive:
@tlaverdure yes. However do make sure sentry/sentry is on 2.1.3 there was another issue that effected cases where “exit” was used (which artisan commands like the worker do).

Otherwise please open a new issue.

tlaverdure:
@stayallive yes I'm on 2.1.3 and I'm receiving errors when the app is running via http but the queued jobs don't seem to be sending errors. Is that supported?

stayallive:
@tlaverdure, it should. However we have not yet tested on Vapor so there might be an Vapor related issue. Please open a new issue so we can look into it!

tlaverdure:
Ok going to do some more testing. The error I was expecting was more so in the configuration of the app, so it may not have been in a job. It looks like this package mainly listens to queue events to gather errors while running queued jobs. I'll open a new issues if errors in jobs are not being captured.

PR #273, opened 2019-09-19
Updates for sentry/sentry 2.2

stayallive:
This PR is a preperation for `sentry/sentry` version 2.2 which deprecates the use of the `Hub` as a singleton.

### DIFF
--- a/src/Sentry/Laravel/EventHandler.php
+++ b/src/Sentry/Laravel/EventHandler.php
@@ -16,7 +16,6 @@ use Illuminate\Routing\Route;
 use RuntimeException;
 use Sentry\State\Scope;
 use Sentry\Breadcrumb;
-use Sentry\State\Hub;
 
 class EventHandler
 {
@@ -115,7 +114,7 @@ class EventHandler
             Integration::flushEvents();
 
             // We have added a scope when a job starts processing
-            Hub::getCurrent()->popScope();
+            Integration::getCurrentHub()->popScope();
         });
 
         foreach (static::$queueEventHandlerMap as $eventName => $handler) {
@@ -283,7 +282,7 @@ class EventHandler
     protected function queueJobProcessingHandler(JobProcessing $event)
     {
         // When a job starts, we want to push a new scope
-        Hub::getCurrent()->pushScope();
+        Integration::getCurrentHub()->pushScope();
 
         $job = [
             'job' => $event->job->getName(),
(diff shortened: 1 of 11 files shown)


composer.json
```diff
@@ -14,7 +14,8 @@
   "require": {
     "php": "^7.1",
     "illuminate/support": "5.0 - 5.8 | ^6.0",
-    "sentry/sdk": "^2.0"
+    "sentry/sdk": "^2.0",
+    "sentry/sentry": "dev-develop@dev"
```
Jean85: I would advise against doing this in the code. For `sentry-symfony`, I was thinking about adding a dedicated job in the CI that would programmatically do the same. Maybe you can do it too?

src/Sentry/Laravel/EventHandler.php
```diff
@@ -115,7 +115,7 @@ public function subscribeQueueEvents(QueueManager $queue)
             Integration::flushEvents();
 
             // We have added a scope when a job starts processing
-            Hub::getCurrent()->popScope();
+            SentrySdk::getCurrentHub()->popScope();
```
Jean85: This approach will require contraining `sentry-php` to `^2.2`; I thought @HazAT preferred the code-based approach, did I got it wrong?

composer.json
stayallive: This is why the PR is a draft 😉 This is not meant to go into the actual PR but for now the only way to pull in the correct code I believe.

src/Sentry/Laravel/EventHandler.php
stayallive: `code-based approach`? What did I miss?

Jean85: The code-based approach would be using `if (class_exists(SentrySdk::class)) { // ... }` or something equivalent.

HazAT: I think we need to do it since we didn't lock the version to minor meaning it could happen that someone updates the core without laravel, so I would prefer the gatekeeper
`if (class_exists(SentrySdk::class)) {` with a comment to remove at some point in the future.

stayallive: Hmm. Good point. Okay 👌

composer.json
stayallive: Added the develop branch to the test matrix too.

src/Sentry/Laravel/EventHandler.php
stayallive: Updated to do a class exists check with plenty of comments about it's removal.

PR #276, opened 2019-09-30
Fix 2.2 compatibility

LasseRafn:
I think the author of the original missed a return statement, because right now its just calling the deprecated method AFTER the new one.


stayallive:
🤦‍♂ Thank you for taking the time to fix this...

PR #278, opened 2019-10-01
Ensure that captured messages are fingerprinted based on the message

MattyRad:
Feature request: Captured messages should have the option to be fingerprinted based on the PSR-3 (https://www.php-fig.org/psr/psr-3/) message.

Hi. Maybe someone can point me in the right direction, but the edited code I have here is an example of an implementation of the feature request, which solves my issue.

We enjoy being able to log messages (*not* exceptions, necessarily) to sentry. However, since every monolog record is unique (based on the time and context e.g. https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/ErrorLogHandler.php#L67), sentry only reports 1 event *per* log message. Needless to say, it's pretty hard to accurately gauge the frequency of each log statement. Fingerprinting seems like the exact feature to avoid this issue, but I don't see an easy way to do that.

What are people's thoughts on adding a feature like this?

### DIFF
--- a/src/Sentry/Laravel/SentryHandler.php
+++ b/src/Sentry/Laravel/SentryHandler.php
@@ -11,6 +11,7 @@ use Sentry\Event;
 use Sentry\Severity;
 use Sentry\State\Hub;
 use Sentry\State\Scope;
+use function Sentry\configureScope;
 
 class SentryHandler extends AbstractProcessingHandler
 {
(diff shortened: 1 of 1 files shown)


MattyRad:
Ugh, I forgot I manually changed this line for the reference value to work. https://github.com/getsentry/sentry-php/blob/master/src/State/Hub.php#L98

```php
        $scope = $this->getScope();

        $callback($scope);
```

I'll work on a better solution that doesn't involve changes to sentry-php, brb heh.

PR #279, opened 2019-10-01
Ensure that captured messages are fingerprinted based on the message

MattyRad:
Original: https://github.com/getsentry/sentry-laravel/pull/278 there's a configure scope function on the hub, I didn't notice that.

Feature request: Captured messages should have the option to be fingerprinted based on the PSR-3 (https://www.php-fig.org/psr/psr-3/) message.

Hi. Maybe someone can point me in the right direction, but the edited code I have here is an example of an implementation of the feature request, which solves my issue.

We enjoy being able to log messages (not exceptions, necessarily, because exceptions are grouped by the stacktrace) to sentry. However, since every monolog record is unique (based on the time and context e.g. https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/ErrorLogHandler.php#L67), sentry only reports 1 event per log message. Needless to say, it's pretty hard to accurately gauge the frequency of each log statement. Fingerprinting seems like the exact feature to avoid this issue, but I don't see an easy way to do that.

What are people's thoughts on adding a feature like this?

### DIFF
--- a/src/Sentry/Laravel/SentryHandler.php
+++ b/src/Sentry/Laravel/SentryHandler.php
@@ -203,6 +203,9 @@ class SentryHandler extends AbstractProcessingHandler
                 if (isset($record['context']['exception']) && $record['context']['exception'] instanceof \Throwable) {
                     $this->hub->captureException($record['context']['exception']);
                 } else {
+                    $this->hub->configureScope(function (Scope $scope) use ($record) {
+                        $scope->setFingerprint([$record['message']]);
+                    });
                     $this->hub->captureMessage($record['formatted'], $this->getLogLevel($record['level']));
                 }
             }


MattyRad:
Ok, I actually missed where the fingerprint can get set through the context. https://github.com/getsentry/sentry-laravel/blob/master/src/Sentry/Laravel/SentryHandler.php#L168-L171 Is that the sole way this repo intends for fingerprints to get set? Because I still think it's worth discussing some options to fingerprint based on the record's message. I'd prefer not to have to extract the message to a variable so it can be re-used through the record's message and the fingerprint both:

```php
$message = 'This is the message';
$log->critical($message, ['fingerprint' => $message]);
```

Either way, being able to set valuable sentry data like tags and fingerprints through the record's context is probably worth putting in the docs.

MattyRad:
Ah, and I just saw this in the docs https://docs.sentry.io/platforms/php/laravel/#user-context. I'm good, I can work this.

I guess the main point I was trying to make is that it makes sense to fingerprint based on the message text, so having an option to do that would be helpful, otherwise it potentially takes some legwork for us to either configure scope and/or set fingerprints in the context.

Sorry for the noise, thanks!

stayallive:
Sentry.io currently has `Grouping settings` in beta (it's in the project settings), I am not 100% sure if everyone sees this feature but this allows setting `Server Side Fingerprinting` which might be what you want to fingerprint based on the message without any code changes for a complete project.

However this is not for everyone since the context around an exception is in most cases as important as the message and the Sentry server is pretty intelligent about grouping the correct exceptions together.

It's highly unlikely we will adopt a change like this since this is massively breaking and not how the grouping is intended to function. I assume you want this because you are using Sentry for logs which is not exactly how Sentry is intended to be used, see: https://sentry.io/vs/logging/.

Hope this helps a bit, feel free to open an issue if you want some clarification on why things are the way they are.

stayallive:
Also, I almost forgot to mention we have a `before_send` hook that allows you to change the Event data before it's being sent which also allows you to update the fingerprint for every event the Sentry client is sending to Sentry.

> https://docs.sentry.io/platforms/php/#filter-events--custom-logic

PR #280, opened 2019-10-04
Do not register default error and fatal listeners and let Laravel bubble these up

### LINKED ISSUE #254: ErrorException: Warning: PDO::prepare(): MySQL server has gone away
There 's a few tickets in the Laravel repositories, but it looks like the reason is an upgrade of the Sentry library to 1.0

Basically we all see `ErrorException Warning: PDO::prepare(): MySQL server has gone away` reports in Sentry, while it supposed to be caught by Laravel, and handle it.

But since the 1.0 update of Sentry library, it somehow bubbles up :)

https://github.com/laravel/horizon/issues/583
https://github.com/laravel/framework/issues/28920

### LINKED ISSUE #261: Two Sentry issues for one exception
Ever since upgrading from `1.0.0-beta2` to `1.1.0` I am sometimes seeing two issues for a single exception, one prefaced with `Notice:`. This was one I created:

image

This could be happening every time, but I'm not sure as I am 2 days from launch. Is this from upgrading or something on my end?

At the time of upgrade I also stopped receiving emails for Laravel issues (but still get JS emails). I have checked my settings (that I didn't ever change) and they look correct to alert me. Anyone else?

stayallive:
Like in Symfony we de-register the error and fatal error listeners so Laravel can handle those and bubble them up through the default [Laravel error handler](https://github.com/laravel/laravel/blob/master/app/Exceptions/Handler.php#L29-L38).

This change prevents duplicate issues (on errors and fatal errors) and issues that arise because they are not filtered out by the frameworks inner workings (mysql warnings silenced by the ORM).

Fixes #261
Fixes #254

### DIFF
--- a/src/Sentry/Laravel/ServiceProvider.php
+++ b/src/Sentry/Laravel/ServiceProvider.php
@@ -7,7 +7,7 @@ use Sentry\ClientBuilder;
 use Sentry\State\HubInterface;
 use Illuminate\Log\LogManager;
 use Laravel\Lumen\Application as Lumen;
-use Sentry\Integration\IntegrationInterface;
+use Sentry\Integration as SdkIntegration;
 use Illuminate\Foundation\Application as Laravel;
 use Illuminate\Support\ServiceProvider as IlluminateServiceProvider;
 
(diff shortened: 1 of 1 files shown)


Christophvh:
Any ETA on this? Since the error spam is crazy the last couple of releases.

PR #290, opened 2019-10-24
Allow for decorating the client builder

### LINKED ISSUE #289: Increase displayed depth of context's arrays
Just noticed I can't access useful information here

array-sentry2

Apparently, there is a way to increase depth in [php sdk manually](https://github.com/getsentry/sentry-php/issues/378#issuecomment-533318398), but does the laravel package handle this as an option ?

Thanks for the help

stayallive:
This allows the user to add this snippet to for example their `AppServiceProvider`s `register` method to set a custom serializer or do other tweaks to the `ClientBuilderInterface;` implementation or swap it out for their own builder interely.

In this example we increase `maxDepth` to 5 in for the default serializer.

```php
use Sentry\Serializer\Serializer;
use Sentry\ClientBuilderInterface;

$this->app->extend(ClientBuilderInterface::class, function (ClientBuilderInterface $clientBuilder) {
    $clientBuilder->setSerializer(new Serializer($clientBuilder->getOptions(), 5));
    
    return $clientBuilder;
});
```

Resolves #289.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Allow decorating the `ClientBuilderInterface` from the `register` method of a Service Provider (#290)
+
 ## 1.4.1
 
 - Fix default Monolog logger level being invalid when using the Log channel (#287)
--- a/src/Sentry/Laravel/ServiceProvider.php
+++ b/src/Sentry/Laravel/ServiceProvider.php
@@ -2,11 +2,11 @@
 
 namespace Sentry\Laravel;
 
-use Sentry\Options;
 use Sentry\State\Hub;
 use Sentry\ClientBuilder;
 use Sentry\State\HubInterface;
 use Illuminate\Log\LogManager;
+use Sentry\ClientBuilderInterface;
 use Laravel\Lumen\Application as Lumen;
 use Sentry\Integration as SdkIntegration;
 use Illuminate\Foundation\Application as Laravel;
@@ -96,7 +96,7 @@ class ServiceProvider extends IlluminateServiceProvider
      */
     protected function configureAndRegisterClient(): void
     {
-        $this->app->bind(ClientBuilder::class, function () {
+        $this->app->bind(ClientBuilderInterface::class, function () {
             $basePath = base_path();
             $userConfig = $this->getUserConfig();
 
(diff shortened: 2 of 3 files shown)


smknstd:
@stayallive as it has been released, did you plan to mention this in the docs ?

stayallive:
@smknstd, yes. It’s taking a little time though. I hope for now the description of the PR will do until the docs can be updated, if you need any clarification let me know because that might benefit the documentation changes later on.

smknstd:
IMO Laravel section in the docs should mention context's additional data. Of course the job belongs more to [laravel](https://laravel.com/docs/5.8/logging#writing-log-messages) and [PSR3](https://www.php-fig.org/psr/psr-3/#13-context), but I think it's an important feature and I'm pretty sure it could help to mention it explicitly. And then it could mention more about it like: 
- the [size limit](https://docs.sentry.io/development/sdk-dev/data-handling/#variable-size
- max depth set to 3 by default, and the fact that sentry hide all array data when the actual depth is too big. 

It's really frustrating to discover those behaviors afterward...

Maybe a snippet like this could help:

```
Log::channel('sentry')->error('something',[
  "a_string" => "value",
  "an_array" => [
     "key1" => "value",
     "key2" => "value",
     "key3" => "value",
     "key4" => "value",
   ]
]);
```

mfn:
I tried this and it "sort of" works, or at least, not in the way I expected it:
- I had to use both `->setSerializer` and `->setRepresentationSerializer`
- the UI change I got went from
  image
- to
  image
I thought I would also get back more collapsing functionality ;)

I wasn't sure if it's worth opening a dedicated issue, maybe that's simply the way it works but I found nothing in the docs regarding the `setRepresentationSeralize` at https://docs.sentry.io/platforms/php/guides/laravel/usage/#customization

thanks!

stayallive:
@HazAT do you happen to know (or who to ask) if there is a server side limit to how many nested fields are shown / collapsable from the sentry.io interface?

Because it looks like there is more data in @mfn screenshots but de UI decided to ignore it possibly.

Jean85:
@stayallive it seems more to me that he changed how the serializer works in the wrong way. 

@mfn there's a specific `$maxDepth` constructor argument for the serializer that decides how much to go deep while serializing something like that, and the default value is `3`. Upping that value changes the behaviour as desired.

mfn:
@stayallive @Jean85 
Why did I not include the code 🤦 I was reporting this based on this code, apologies for not posting this immediately:
```php
        $this->app->extend(ClientBuilderInterface::class, function (ClientBuilderInterface $clientBuilder): ClientBuilderInterface {
            $clientBuilder->setSerializer(new Serializer($clientBuilder->getOptions(), 5));
            $clientBuilder->setRepresentationSerializer(new RepresentationSerializer($clientBuilder->getOptions(), 5));

            return $clientBuilder;
        });
```

mfn:
sorry for the ping @stayallive @Jean85 but can you check my last comment? ☝️
Or should I create a new issue?

stayallive:
@mfn, sorry I missed your response, it looks like you are doing it right but my assumption is that this is a UI "problem" and that the Sentry UI might not allow more nesting to show.

Could you share (the relevant section) of the event JSON to make sure it does come across correctly nested?

And yeah it's probably a good idea to move this to a new issue so we don't discuss this on a pretty unrelated PR :)

PR #296, opened 2019-11-01
Allow any log level to create breadcrumbs

### LINKED ISSUE #297: Breadcrumb for all MessageLogged events
Creating a breadcrumb with an invalid log level causes an exception. Currenly the EventHandler does not parse or convert the received log level to one of the supported breadcrumb levels, resulting in the log being missing in the breadcrumbs.

### Test code

```php
Route::get('/debug-sentry', function () {
    Log::debug('Debug log {debug_context}', ['debug_context' => 1]);
    Log::info('Info log {debug_context}', ['debug_context' => 2]);
    Log::notice('Notice log {debug_context}', ['debug_context' => 3]);
    Log::warning('Warning log {debug_context}', ['debug_context' => 4]);
    Log::error('Error log {debug_context}', ['debug_context' => 5]);
    Log::critical('Critical log {debug_context}', ['debug_context' => 6]);
    Log::alert('Alert log {debug_context}', ['debug_context' => 7]);
    Log::emergency('Emergency log {debug_context}', ['debug_context' => 8]);

    throw new Exception('My first Sentry error!');
});
```

### Current behavior

image

As you can see the `notice`, `alert` and `emergency` logs are missing in the Sentry report.

### Expected behavior

image

All PSR3 log levels are present event though Sentry does not render all it's breadcrumb levels distinctly (I would expect fatal to be more red than error at least :smile:, but this is a different issue as well as using the user type for logs)

### Solution

I have created a pull request (#296) to resolve this issue by adding a helper method to parse any log level to one of the supported breadcrumb levels, defaulting to the info level.

TheLevti:
Creating a breadcrumb with an invalid log level would cause an exception, resulting in the log being missing in the breadcrumbs. This pull request adds a log level adapter method, which translates common log levels to suitable breadcrumb levels, defaulting to the info level.

Fixes #297

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
   "require": {
     "php": "^7.1",
     "illuminate/support": "5.0 - 5.8 | ^6.0",
-    "sentry/sdk": "^2.0.4"
+    "sentry/sdk": "^2.0"
   },
   "require-dev": {
     "phpunit/phpunit": "^8.0",
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
   "require": {
     "php": "^7.1",
     "illuminate/support": "5.0 - 5.8 | ^6.0",
-    "sentry/sdk": "^2.0"
+    "sentry/sdk": "^2.0.4"
   },
   "require-dev": {
     "phpunit/phpunit": "^8.0",
--- a/src/Sentry/Laravel/EventHandler.php
+++ b/src/Sentry/Laravel/EventHandler.php
@@ -3,19 +3,19 @@
 namespace Sentry\Laravel;
 
 use Exception;
-use Illuminate\Log\Events\MessageLogged;
 use Illuminate\Auth\Events\Authenticated;
+use Illuminate\Console\Events\CommandFinished;
+use Illuminate\Console\Events\CommandStarting;
+use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Database\Events\QueryExecuted;
+use Illuminate\Log\Events\MessageLogged;
 use Illuminate\Queue\Events\JobProcessing;
 use Illuminate\Queue\QueueManager;
 use Illuminate\Routing\Events\RouteMatched;
-use Illuminate\Contracts\Events\Dispatcher;
-use Illuminate\Database\Events\QueryExecuted;
-use Illuminate\Console\Events\CommandStarting;
-use Illuminate\Console\Events\CommandFinished;
 use Illuminate\Routing\Route;
 use RuntimeException;
-use Sentry\State\Scope;
 use Sentry\Breadcrumb;
+use Sentry\State\Scope;
 
 class EventHandler
 {
@@ -240,13 +240,7 @@ class EventHandler
      */
     protected function logHandler($level, $message, $context)
     {
-        Integration::addBreadcrumb(new Breadcrumb(
-            $level,
-            Breadcrumb::TYPE_USER,
-            'log.' . $level,
-            $message,
-            empty($context) ? [] : ['params' => $context]
-        ));
+        $this->addLogBreadcrumb($level, $message, is_array($context) ? $context : []);
     }
 
     /**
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -14,7 +14,7 @@
   "require": {
     "php": "^7.1",
     "illuminate/support": "5.0 - 5.8 | ^6.0",
-    "sentry/sdk": "^2.0"
+    "sentry/sdk": "^2.0.4"
```
stayallive: Is this needed in the context of this PR?

src/Sentry/Laravel/EventHandler.php
```diff
@@ -240,13 +240,7 @@ private function addQueryBreadcrumb($query, $bindings, $time, $connectionName)
...
     protected function logHandler($level, $message, $context)
     {
-        Integration::addBreadcrumb(new Breadcrumb(
-            $level,
-            Breadcrumb::TYPE_USER,
-            'log.' . $level,
-            $message,
-            empty($context) ? [] : ['params' => $context]
-        ));
+        $this->addLogBreadcrumb($level, $message, is_array($context) ? $context : []);
```
stayallive:
```suggestion
        $this->addLogBreadcrumb($level, $message, $context);
```

Not better to just leave it like this and let `addLogBreadcrumb` to decide what to do with the input?

src/Sentry/Laravel/EventHandler.php
```diff
@@ -255,16 +249,56 @@ protected function logHandler($level, $message, $context)
...
+     * @param  string $level   Log level. May be any standard.
+     * @param  string $message Log messsage.
+     * @param  array  $context Log context.
+     * @return void
+     */
+    private function addLogBreadcrumb(string $level, string $message, array $context = []): void
     {
         Integration::addBreadcrumb(new Breadcrumb(
-            $logEntry->level,
+            $this->logLvlToBreadcrumbLvl($level),
```
stayallive:
```suggestion
            $this->logLevelToBreadcrumbLevel($level),
```

No need to shorten this IMHO.

src/Sentry/Laravel/EventHandler.php
```diff
@@ -255,16 +249,56 @@ protected function logHandler($level, $message, $context)
...
+                return Breadcrumb::LEVEL_WARNING;
+            case 'error':
+                return Breadcrumb::LEVEL_ERROR;
+            case 'critical':
+            case 'alert':
+            case 'emergency':
+                return Breadcrumb::LEVEL_FATAL;
+            default:
+                return Breadcrumb::LEVEL_INFO;
+        }
```
stayallive: This can be shortened a little to:

```php
        switch (strtolower($level)) {
            case 'debug':
                return Breadcrumb::LEVEL_DEBUG;
            case 'warning':
                return Breadcrumb::LEVEL_WARNING;
            case 'error':
                return Breadcrumb::LEVEL_ERROR;
            case 'critical':
            case 'alert':
            case 'emergency':
                return Breadcrumb::LEVEL_FATAL;
            case 'info':
            case 'notice':
            default:
                return Breadcrumb::LEVEL_INFO;
        }
```

src/Sentry/Laravel/EventHandler.php
```diff
@@ -255,16 +249,56 @@ protected function logHandler($level, $message, $context)
...
         ));
     }
 
+    /**
+     * Translates common log levels to Sentry breadcrumb levels.
+     *
+     * @param string $level Log level. Maybe any standard.
+     * @return string Breadcrumb level.
+     */
+    protected function logLvlToBreadcrumbLvl(string $level): string
```
stayallive:
```suggestion
    protected function logLevelToBreadcrumbLevel(string $level): string
```

src/Sentry/Laravel/EventHandler.php
```diff
@@ -255,16 +249,56 @@ protected function logHandler($level, $message, $context)
...
+        $this->addLogBreadcrumb($logEntry->level, $logEntry->message, $logEntry->context);
+    }
+
+    /**
+     * Helper to add an log breadcrumb.
+     *
+     * @param  string $level   Log level. May be any standard.
+     * @param  string $message Log messsage.
+     * @param  array  $context Log context.
+     * @return void
```
stayallive:
```suggestion
```

src/Sentry/Laravel/EventHandler.php
```diff
@@ -255,16 +249,56 @@ protected function logHandler($level, $message, $context)
...
+            $message,
+            empty($context) ? [] : ['params' => $context]
         ));
     }
 
+    /**
+     * Translates common log levels to Sentry breadcrumb levels.
+     *
+     * @param string $level Log level. Maybe any standard.
+     * @return string Breadcrumb level.
```
stayallive:
```suggestion
     *
     * @return string Breadcrumb level.
```

composer.json
TheLevti: Yes this is required to inlude the fixed/missing breadcrumb level `fatal`. 

See: https://github.com/getsentry/sentry-php/pull/902

src/Sentry/Laravel/EventHandler.php
TheLevti: No, because `addLogBreadcrumb` expects a valid context array. This legacy handler method may receive null as the context though. I would rather prefer that the actual method that adds the breadcrumb can expect valid log data (level, message and context).

Especially, because its legacy.

composer.json
TheLevti: Ok actually I expected that the latest sentry sdk requires the latest sentry-php. But it just requires ^2.1.3 up from ^2.0.1 with this change. Which means that the required change might not be in there if a user explicitly wants something between 2.1.3 and 2.2.2.

TheLevti: We could skip this version bumb until sentry/sdk bumps it's version of sentry-php. What do you suggest?

src/Sentry/Laravel/EventHandler.php
stayallive: Okay agreed.

composer.json
stayallive: I see. It's not optimal but what about we leave the requirement and do this until we can depend on 3.x?

```php
        // @TODO: Once we depend on `sentry-php` 3.0 we can remove this check and use `Breadcrumb::LEVEL_FATAL`.
        $fatal = defined(Breadcrumb::class . '::LEVEL_FATAL')
            ? Breadcrumb::LEVEL_FATAL
            : Breadcrumb::LEVEL_CRITICAL;

        switch (strtolower($level)) {
// ...
            case 'critical':
            case 'alert':
            case 'emergency':
                return $fatal;
// ...
        }
```

stayallive:
FYI: I've rebased this PR because some other conflicting PR's got merged.

composer.json
TheLevti: Sound reasonable. I will add the suggested solution.

stayallive:
Thanks for taking the time to create the PR! Much appreciated! 👍

PR #312, opened 2019-12-12
Replace QueueManager::looping() with  QueueManager::after()

tlaverdure:
When Laravel processes queued jobs the event `Illuminate\Queue\Events\Looping` is only dispatched when the queue worker is running as a daemon. When processing only one job with `php artisan queue:work --once` no exceptions will be captured. This is also true for queued jobs in Laravel Vapor since jobs are processed one at a time in in their own instance of the application.

Using `QueueManager::after()` will provide a more consistent experience for capturing exceptions.


tlaverdure:
@stayallive following up about https://github.com/getsentry/sentry-laravel/pull/271#issuecomment-531337093

tlaverdure:
Any word on this?

stayallive:
Hi @tlaverdure, I'm sorry for the delay. There is an issue with using `after` and it's the reason it isn't used instead of `looping`.

https://github.com/laravel/framework/blob/4e33586e6cf6baf3a857e8ec6bd6de2ae4f93ad9/src/Illuminate/Queue/Worker.php#L356

At the snippet above is where the `after` is called, however this is never called after a job fails. So that's why looping was used. However I see that that is also not correct since as you pointed out it only runs for deamons.

So I've changed the code to utilize more events that should be fired everywhere and always handling both successful jobs and jobs with exceptions (and might also improve the situation when using Vapor for example).

Please take a look at: #318

tlaverdure:
Sounds good. Thanks for taking a look into this. #318 looks like a great solution.

PR #322, opened 2020-01-23
Fix queue events with missing handler suffix

besanek:
Hello,

we encountered a queue problem after upgrading to version 1.6.0.
```
In EventHandler.php line 154:
   Missing event handler: queueWorkerStoppinghandler
```
After downgrade to 1.5.0 everything is correct again.
It seems that the EventHandler::_call() checks for the presence queueWorkerStoppinghandler however, the name is queueWorkerStopping.

Solves #323


stayallive:
Great testing on my part, thanks for fixing it!

bradders:
Thanks so much for this fix! Only noticed this happening when my Sentry account got frozen from too many errors popping up! 😭

PR #340, opened 2020-04-06
Add `send_default_pii` to published config file

ankurk91:
Hi,

This config change will send logged-in user id along with error.
This should be enabled by default.

If you guys don't want to enable this default, let's just keep this property in config file so that we (developers) can enable it without going into docs anytime.

Thanks.

### DIFF
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -21,5 +21,7 @@ return [
         'queue_info' => true,
     ],
 
-    'send_default_pii' => true,
+    // @see: https://docs.sentry.io/error-reporting/configuration/?platform=php#send-default-pii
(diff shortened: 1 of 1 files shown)


config/sentry.php
```diff
@@ -21,4 +21,5 @@
         'queue_info' => true,
     ],
 
+    'send_default_pii' => true,
```
stayallive, line 24:
```suggestion
    // @see: https://docs.sentry.io/error-reporting/configuration/?platform=php#send-default-pii
    'send_default_pii' => false,

```

ankurk91:
@stayallive 
I have accepted you changes, please check.

PR #351, opened 2020-05-28
Fix job event and scope handling

### LINKED ISSUE #342: Scope data in queue jobs are lost
Version: **v1.7.0**

Problem: **Adding extras/tags in a queue-jobs handle() method are not being pushed to sentry**

## How to reproduce:

I created a TestSentryJob like this:

```php
public function handle()
{
    dump("TestSentryJob is running..");

    configureScope(function (Scope $scope)
    {
        $scope->setTag("testtag","test");
        $scope->setExtra("testid","id");
    });

    throw new \Exception("test");
}
```

If I run that in Sync mode it works, **if I run it via the Database Queue Driver, the extra+tag is not being pushed.**

I further added some logging to the following files:

`./sentry-laravel/src/Sentry/Laravel/EventHandler.php:beforeQueuedJob()`
`./sentry-laravel/src/Sentry/Laravel/EventHandler.php:afterQueuedJob()`

```php
private function beforeQueuedJob()
{
    // When a job starts, we want to push a new scope
    dump("We push a new Scope");
    SentrySdk::getCurrentHub()->pushScope();
}

private function afterQueuedJob()
{
    dump("We pop the scope");
    // Flush any and all events that were possibly generated by queue jobs
    Integration::flushEvents();

    // We have added a scope when the job started processing
    SentrySdk::getCurrentHub()->popScope();
}
```

`./sentry/sentry/src/State/Hub.php:captureException`

```php
public function captureException(\Throwable $exception): ?string
{
    $client = $this->getClient();

    dump("We capture the Exception");

    if (null !== $client) {
        return $this->lastEventId = $client->captureException($exception, $this->getScope());
    }

    return null;
}
```

which results in the following output:

```
"We push a new Scope"
[2020-04-11 18:53:28][827] Processing: App\Jobs\TestSentryJob
"TestSentryJob is running.."
[2020-04-11 18:53:28][827] Failed:     App\Jobs\TestSentryJob
"We pop the scope"
"We capture the Exception"
```

It looks like the scope is getting killed **before** its captured

(If I am removing the pop/push in the before and/or after QueuedJob Events, the correct scope gets pushed along with the error)

## Possible Solutions:

As I'm not very familiar with your codebase I'm not sure why you need push/pop the Scope for every queue QueueJob (I imagine that is for the Daemon mode which won't kill the current process.) - my thought was to (only) start a fresh Scope with every job 

Might be worth looking into the `Looping` Event which gets fired Laravel decides if the Daemon should keep on running 🤷‍♀️

stayallive:
This should fix #342.

There is only 1 edge case left. When running the `sync` driver and nesting queued jobs (dispatching a queue job from a dispatched job) only the scope data of the last job started is captured, this is _only_ and issue on the `sync` driver (which is not a production driver, but a development one) so I'm personally fine with this.

This should work great bot in daemon mode and just `queue:work` mode.

I have tried a whole evening to test this, but there is no way without actually running an app and a queue worker to emulate it correctly so I do not want to add half-assed tests, this might be revisited later. I validated this works by just trying to break it using a scratch Laravel app and it seems to be working.


arminulrich:
Awesome! for me it solved the issue and I was able to attach extras 👍

stayallive:
Thanks for verifying, reporting and being patient <3 

I'll try to get this released next week 👍

troyharvey:
🙏  Love it. Thanks gentlemen.

PR #374, opened 2020-09-03
Laravel 8 support

### LINKED ISSUE #371: Laravel 8 support
Laravel 8 support for the Sentry Laravel integration will be released on the day (currently planned for 8 september 2020) that Laravel 8 is released.

No need to create PR's or issues for this, it is on the radar and we will make sure Laravel 8 support will be there on release day 💪

stayallive:
Fixes #371.

This should be released in 2.x before Sept 8th (Laravel 8 release date).

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,7 +1,9 @@
 <?php
+
 $finder = PhpCsFixer\Finder::create()
     ->in(__DIR__ . '/src')
 ;
+
 return PhpCsFixer\Config::create()
     ->setRules([
         '@PSR2' => true,
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,20 +22,18 @@ env:
     - LARAVEL=^7.0 TESTBENCH=5.1.* PHPUNIT=8.4.* SENTRY=^2.3
 
     # We add one more test using the next version of Laravel which only support PHP ^7.3 (Laravel requirement)
-    - LARAVEL=8.x-dev@dev TESTBENCH=6.0.x-dev@dev PHPUNIT=8.4.* SENTRY=^2.3 COMPOSER_STABILITY=dev
+    - LARAVEL=8.x-dev@dev TESTBENCH=^6.0 PHPUNIT=8.4.* SENTRY=^2.3 COMPOSER_STABILITY=dev
 
 matrix:
   fast_finish: true
   allow_failures:
-    - php: 7.2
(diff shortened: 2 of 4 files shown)


georgeboot:
Note that `sentry/sdk` requires `php-http/guzzle6-adapter` which is incompatible with Laravel 8 (requires Guzzle 7).

georgeboot:
@stayallive this should not have been merged as is.

Laravel 8 required guzzle 7 and that conflicts with the guzzle 6 adapter required by https://github.com/getsentry/sentry-php-sdk

So Laravel 8 is not yet supported.

stayallive:
Laravel does not (directly) requires Guzzle 7 (https://github.com/laravel/framework/blob/8.x/composer.json), did I miss something?

I've also tested on a clean Laravel 8 application before release which installed without problems.

mfn:
I'm trying to figure this out too, wonder if https://github.com/laravel/laravel/pull/5381 causes the problem.

aka new installations of L8 have this in, so effectively requiring Guzzle 7

georgeboot:
Jup. The upgrade guide mentiones Guuzle ^7.0.1: https://laravel.com/docs/8.x/upgrade

stayallive:
That's unfortunate, until we support Guzzle 7 this is your best option, I'll see if we can find a shorter way and add this to the documentation:

```
{
    "require": {
        "sentry/sentry-laravel": "^1.9",
        "sentry/sentry": "^2.4",
        "http-interop/http-factory-guzzle": "^1.0",
        "php-http/curl-client": "^2.0",
        "guzzlehttp/guzzle": "^7.0"
    },
    "replace": {
        "sentry/sdk": "*"
    }
}
```

mfn:
Maybe chime in on https://github.com/laravel/laravel/pull/5381 , maybe we can get ^6 there. I haven't yet seen a strict reason as to why it's required, and for all intents and purposes, on a high level they're compatible.

stayallive:
Since Guzzle 7 will be used more and more and there is no easy way support both (when using httplug) I've opted to see if we can replace Guzzle for the Sentry SDK so we can't conflict or force some version like 6 which people want to upgrade to. 

https://github.com/getsentry/sentry-php-sdk/pull/7

For quick upgraders there is an workaround in the pinned issue but otherwise know we are trying to find a solution for it (it does not only affect Laravel). I must admit I regret missing this change which caught me off guard, I was sure we were prepared.

PR #387, opened 2020-09-28
Improve tracing

stayallive:
- [x] Nest `view.render` spans
- [x] Nest `sql.query` spans
- [x] Add 2 child spans to `http.server`
  - [x] `app.bootstrap` the Laravel bootstrap process
  - [x] `app.handle` (please suggest a better name) which will have all other spans that are generated by the user application as childs
- [x] Only boot the tracing provider if there was a hint of a DSN (like the Sentry SDK itself)
- [x] Move all tracing related events to it's own event handler

Current state:

image

### DIFF
--- a/src/Sentry/Laravel/Tracing/Middleware.php
+++ b/src/Sentry/Laravel/Tracing/Middleware.php
@@ -4,6 +4,7 @@ namespace Sentry\Laravel\Tracing;
 
 use Closure;
 use Illuminate\Http\Request;
+use Illuminate\Http\Response;
 use Sentry\SentrySdk;
 use Sentry\State\Hub;
 use Sentry\Tracing\SpanContext;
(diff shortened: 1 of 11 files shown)


HazAT:
Like the direction 👍

src/Sentry/Laravel/Tracing/Middleware.php
```diff
@@ -46,6 +53,13 @@ public function handle($request, Closure $next)
     public function terminate($request, $response): void
     {
         if ($this->transaction !== null && app()->bound('sentry')) {
+            if ($this->appSpan !== null) {
+                $this->appSpan->finish();
+            }
+
+            // Make sure we set the span in the Sentry SDK to the transaction
```
HazAT, line 61: Maybe write here also:

If the `transaction` is not on the scope during finish, the trace.context is wrong

HazAT (approved):
Cool stuff, Changelog and we can release ^^

HazAT (approved):

PR #390, opened 2020-09-30
Disable tracing on empty `traces_sample_rate`

stayallive:
Noticed we can probably disable the tracing completely if the `traces_sample_rate` is `0.0` since there would be no chance to send a trace to Sentry which would negate any performance impact the tracing code could have on an application.

Looking at this code do we want to collect the trace data anyway even if we already know the sample rate is 0 and thus the trace will not be sent or is the `traces_sampler` option supposed to be able to override that later? Because then this PR might not be good...

Bottom line, if the `traces_sample_rate === 0.0` should we even collect tracing data?
Yes = Merge, No = Close 😉

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Improve performance tracing by nesting `view.render` spans and adding a `app.handle` span showing how long the actual application code runs after Laravel bootstrapping (#387)
+- Disable tracing on empty `traces_sample_rate` (#390)
 
 ## 2.0.0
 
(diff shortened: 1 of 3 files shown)


HazAT:
Tricky, PHP is a bit unique in that regard since it basically has no persistence in any way between requests. So it should be ok to do it. But let's wait a bit until we gathered more feedback to release this.

stayallive:
After consideration and looking at real world performance data I have decided to postpone this until actually needed since currently the impact is close to 0 and the impact with all the ways tracing can be sampled it's not worth it right now.

PR #403, opened 2020-10-20
Skip performance tracing in unit tests

leo108:
Since [startTransaction](https://github.com/getsentry/sentry-laravel/blob/master/src/Sentry/Laravel/Tracing/Middleware.php#L73) is using typehint and class `Sentry\State\Hub` is declared as final, we are not able to mock `app('sentry')` in unit tests, I don't think we should trace performance during unit tests.

### DIFF
--- a/src/Sentry/Laravel/Tracing/Middleware.php
+++ b/src/Sentry/Laravel/Tracing/Middleware.php
@@ -36,7 +36,7 @@ class Middleware
      */
     public function handle($request, Closure $next)
     {
-        if (app()->bound('sentry')) {
+        if (app()->bound('sentry') && !app()->runningUnitTests()) {
             $this->startTransaction($request, app('sentry'));
         }
 


leo108:
@stayallive Do you have a chance to check this?

stayallive:
Hi @leo108, sorry about the late response, I'm thinking on how we could solve this better since wat you are proposing will not work since we still support Laravel 5.0 which does not have the `runningUnitTests` method which was introduced in 5.6.

Wouldn't changing the typehint to `HubInterface` solve the issue too, you should be able to mock an interface?

However, you should just set `SENTRY_LARAVEL_DSN` to `null` (for example here: https://github.com/laravel/laravel/blob/master/phpunit.xml#L20-L30) in your Unit tests and not deal with any Sentry specific code I would say. If you are integration testing with Sentry you can wrap the code with `app()->bound('sentry')` to prevent calling Sentry code.

stayallive:
I am closing this issue because of inactivity, this is also solved by setting `SENTRY_LARAVEL_DSN` to null and/or set the `tracing_sample_rate` to `0` when running tests.

PR #405, opened 2020-10-23
Use as-is context in breadcrumbs

Wirone:
Decorating passed context with `['params' => $context]` does not make sense and has negative impact on readability of provided data.

Before:
image

After:
image

FYI: I didn't run tests because I've edited this with Github UI (after testing locally in project), but it should pass.


stayallive (approved):
So this took a second... thanks, this makes perfect sense 👍

PR #413, opened 2020-11-06
Fix variable duplication

baijunyao:


stayallive (approved):
Thanks, this happens in another place too, will fix that also 😄

Issue #410, opened 2020-11-05
Tracing for Lumen?


giggsey:
Hi,

Does the Tracing support Lumen?

Registering that extra Service Provider doesn't seem to work.

addeeandra:
Had the same issue, after that I added `Tracing\Middleware::class` in `bootstrap/app.php`:

```php
$app->middleware([
    \Sentry\Laravel\Tracing\Middleware::class,
    \Fideloper\Proxy\TrustProxies::class,
    ...
]);
```

It's seems not resolving :
`Call to undefined method Laravel\Lumen\Application::booted() in Middleware.php line 104`

Then I checked the Middleware at line 104, I think this lines of code cause the issue :

```php
if (!$this->addBootTimeSpans()) {
    // @TODO: We might want to move this together with the `RouteMatches` listener to some central place and or do this from the `EventHandler`
    app()->booted(function () use ($request, $fallbackTime): void {
        $spanContextStart = new SpanContext();
        $spanContextStart->setOp('app.bootstrap');
        $spanContextStart->setStartTimestamp(defined('LARAVEL_START') ? LARAVEL_START : $request->server('REQUEST_TIME_FLOAT', $fallbackTime));
        $spanContextStart->setEndTimestamp(microtime(true));
        $this->transaction->startChild($spanContextStart);

        $appContextStart = new SpanContext();
        $appContextStart->setOp('app.handle');
        $appContextStart->setStartTimestamp(microtime(true));
        $this->appSpan = $this->transaction->startChild($appContextStart);

        SentrySdk::getCurrentHub()->setSpan($this->appSpan);
    });
}
```

addeeandra:
Well, I did some work around.

First, by defining `LARAVEL_START`, `SENTRY_AUTOLOAD`, and `SENTRY_BOOSTRAP` somewhere at my code i.e. `define(LARAVEL_START, microtime(true))` in `bootstrap/app.php`, it will make the `$this->addBootTimeSpans()` returning true and it will just pass by the `app()->booted($closure)` statement.

It will create new SpanContext from it :

```php
// Sentry\Laravel\Tracing\Middleware.php:121

private function addBootTimeSpans(): bool
{
    if (!defined('LARAVEL_START') || !LARAVEL_START) {
        return false;
    }

    if (!defined('SENTRY_AUTOLOAD') || !SENTRY_AUTOLOAD) {
        return false;
    }

[...]

PR #416, opened 2020-11-14
Fix problems when enabling tracing on Laravel Lumen

stayallive:
This fixes problems with tracing for Lumen.

Lumen does not have a `booted` event, so we do have to rely on the user adding the `LARAVEL_START`, `SENTRY_AUTOLOAD` and `SENTRY_BOOTSTRAP` defines to their application files for correct bootstrap instrumentation, but that seems like a documentation issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fix problems when enabling tracing on Laravel Lumen (#416)
 - PHP 8 Support (#431)
 - Bump Sentry SDK to `3.2.*` (#431)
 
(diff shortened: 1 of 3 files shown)


addeeandra:
Hi, I have tried the code and it works fine even in a clean Lumen installation.

It works as expected with and without defining the constants like I mention before ( #410 ) ..

Simply registering `\Tracing\ServiceProvider::class` in `bootstrap/app.php` to enable the Tracing and it works fine :

```php
$app->register(Sentry\Laravel\ServiceProvider::class);
$app->register(\Sentry\Laravel\Tracing\ServiceProvider::class);
...
```

PR #420, opened 2020-11-19
feat: Use fromSentryTrace & Bump SDK

HazAT:


stayallive:
Also requires us to bump the required SDK version to prevent errors once this is released in the PHP SDK?

stayallive (approved):
👍

Issue #424, opened 2020-11-27
PHP 8 Support


vladislavtkachenko:
Hello, do you have any plans to support php 8? 
requires php ^7.2, you php ver 8.0.0

Thank you for your work

stayallive:
Yes we have. You can follow along with our progress here: getsentry/sentry-php#1087.

vladislavtkachenko:
Thank you

bulgariamitko:
When will be sentry PHP 8.0 ready, as I tried today and I got those errors:

```
Warning: stream_wrapper_register(): Protocol symfony:// is already defined. in /home/***/public_html/vendor/symfony/http-client/Response/StreamWrapper.php on line 64

Warning: fopen(): symfony:// wrapper is disabled in the server configuration by allow_url_fopen=0 in /home/**/public_html/vendor/symfony/http-client/Response/StreamWrapper.php on line 74

Warning: stream_wrapper_register(): Protocol symfony:// is already defined. in /home/**/public_html/vendor/symfony/http-client/Response/StreamWrapper.php on line 64

Warning: fopen(symfony://https://****.ingest.sentry.io/api/****/store/): Failed to open stream: no suitable wrapper could be found in /home/***/public_html/vendor/symfony/http-client/Response/StreamWrapper.php on line 74

Fatal error: Uncaught TypeError: round(): Argument #1 ($num) must be of type int|float, string given in /home/**/public_html/***/aconfig/***.php:91 Stack trace: #0 
```

stayallive:
You should be using `2.3.1` of this package and that should work just great.

However it looks like your php configuration has some things disabled that makes the HTTP library we ship unable to sent events to sentry.io: `fopen(): symfony:// wrapper is disabled in the server configuration by allow_url_fopen=`.

You can check out how to change the transport if you cannot use the default Symfony HTTP client: https://docs.sentry.io/platforms/php/guides/laravel/configuration/transports/.

If that still doesn't solve the problem please open up a new issue so we can investigate!

PR #425, opened 2020-11-28
Add PHP 8 support

annejan:
Add PHP 8 support


stayallive:
See #424, it's not this simple and work is underway, a little more patience please.

PR #428, opened 2020-11-30
Support PHP 8

m1guelpf:


stayallive:
Hey @m1guelpf we are working on it but you probably know it's not this simple (see #425) 😉 

See #424 to see that we are blocked by getsentry/sentry-php#1087.

PR #432, opened 2020-12-06
Fix link for send-default-pii in config

NiclasvanEyk:
The old one just redirects to https://docs.sentry.io/platforms/php and does not link to the explanation for send-default-pii


stayallive (approved):
Appreciated, thanks!

PR #441, opened 2021-01-20
Fix generated route name not correctly detected when using prefixes

### LINKED ISSUE #440: Generated route name not correctly detected when using prefixes
It should be `Str::contains(...)` instead of `Str::startsWith(...)` in https://github.com/getsentry/sentry-laravel/blob/c5af38e42f5cd513731613a64a2c6bc8039892a0/src/Sentry/Laravel/Integration.php#L138

Otherwise, you get routes like "prefix.generated::aiusdha" in Sentry.

wfeller:
Use Str::contains instead of Str::startsWith to verify if route name is generated.

Fixes #440

### DIFF
--- a/src/Sentry/Laravel/Integration.php
+++ b/src/Sentry/Laravel/Integration.php
@@ -135,7 +135,7 @@ class Integration implements IntegrationInterface
             // Laravel 7 route caching generates a route names if the user didn't specify one
             // theirselfs to optimize route matching. These route names are useless to the
             // developer so if we encounter a generated route name we discard the value
-            if (Str::startsWith($routeName, 'generated::')) {
+            if (Str::contains($routeName, 'generated::')) {
                 $routeName = null;
             }
 


stayallive:
Hey @wfeller I have been testing this but I don't see this problem occurring...

I am testing on Laravel 8, which version are you using?

When using the route below and running `php artisan route:cache` I get: `generated::FRnHID1MBC9OnhFk` and not `test.generated::FRnHID1MBC9OnhFk`.

```php
Route::group([
    'name' => 'test.',
], function () {
	Route::view('test', 'home');
})
```

wfeller:
We're on Laravel 8.23.1

Try it with this maybe - I think the name argument doesn't do the same thing as prefix:
```php
Route::prefix('prefix.')->group(function () {
    Route::view('test', 'home');
});
```

From my Sentry performance tab:
image

stayallive:
I don't think this does what I think it would do...

```php
Route::prefix('pizza.')->group(function () {
    Route::view('test', 'home');
});
```

This generates the route: `app.develop/pizza./test`, I don't think that is what you intended?

Below generates the route: `app.develop/test` as I would expect.

```php
Route::name('pizza.')->group(function () {
    Route::view('test', 'home');
});
```

Maybe I'm totally missing the boat here and you mean to use the prefix that way but I don't think that's right.

But even with that inspecting my `bootstrap/cache/routes-v7.php` I cannot get Laravel to generate the prefix before the generated name :(

wfeller:
My bad, I looked at my routes file and saw prefix first, thought it had to be it! (we actually use both `prefix` and `as`)

I can reproduce it in a fresh Laravel codebase:
```php
Route::as('prefix.')->group(function () {
    Route::get('test-1', 'Controller@action');
    Route::get('test-2', 'Controller@action');
});
```

This will generate 2 cached routes:
"prefix.": /test-1
"prefix.generated::U282gBpR9Z1vbfVc": /test-2

It doesn't seem possible to reproduce the issue with only 1 unnamed route.

stayallive:
Yep that did it! Sorry about the many questions just wanted to make sure this was needed.

Thank you for your patience, I hope to release this in the next week.

PR #457, opened 2021-03-03
Add event processor that adds exception context to event

stayallive:
Since (very) recently Laravel added the implicit contract the the public `context` method on an exception class returns some extra context information that is added to the logs.

We like to also relay this information to Sentry.

Example (from the Laravel docs):

```php
<?php

namespace App\Exceptions;

use Exception;

class InvalidOrderException extends Exception
{
    // ...

    /**
     * Get the exception's context information.
     *
     * @return array
     */
    public function context()
    {
        return ['order_id' => $this->orderId];
    }
}
```

> https://laravel.com/docs/8.x/errors#exception-log-context

Laravel adds this information to the logs and Sentry adds this information to the `exception_context` key in the extra data.

This looks like this in the Sentry interface:

image

### DIFF
--- a/src/Sentry/Laravel/Integration/ExceptionContextIntegration.php
+++ b/src/Sentry/Laravel/Integration/ExceptionContextIntegration.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Sentry\Laravel\Integration;
+
+use Sentry\Event;
+use Sentry\EventHint;
+use Sentry\State\Scope;
+use Sentry\Integration\IntegrationInterface;
+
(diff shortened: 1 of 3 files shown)


olivernybroe:
Nice 👍 Simple and simple.

Is there any limits to the amount of nesting which can happen on the additional data that people have to be aware of?

stayallive:
Yes, the default amount is 3 in the SDK and the Sentry server also enforces a limit on which it collapses or truncates the data (not sure what the exact limit is there). 

But I would generally advice to keep this to a minimum and not add too much (nested) data if you can help it.

PR #467, opened 2021-03-24
Avoid collision if another package has bound 'sentry'

### LINKED ISSUE #466: Changes from #419 are not compatible with projects that change the BaseServiceProvider::$abstract
### Environment

How do you use Sentry?
sentry.io but this bug occurs with or without a DSN set

Which SDK and version?
PHP and 3.1.5 (whatever `composer require "sentry/sentry-laravel 2.3.1"` delivers)

### Steps to Reproduce

1. composer require sentry/sentry-laravel (2.4.1)
2. add sentry/sentry-laravel to do-not-discover in composer.json
3. create a replacement ServiceProvider that extends BaseServiceProvider and sets BaseServiceProvider::$abstract to something other than 'sentry'

### Expected Result

Middleware stack processes without throwing.

### Actual Result

SetRequestIpMiddleware (added in #419 ) throws because `app('sentry')` is not an instance of `HubInterface`.

KarlCastle:
Resolves #466

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Avoid collision if another package has bound `sentry` in the Laravel container (#467)
+
 ## 2.4.1
 
 - Fix type hints incompatible with Laravel Lumen (#462)
--- a/src/Sentry/Laravel/Facade.php
+++ b/src/Sentry/Laravel/Facade.php
@@ -21,6 +21,6 @@ class Facade extends \Illuminate\Support\Facades\Facade
 {
     protected static function getFacadeAccessor()
     {
-        return 'sentry';
+        return Sentry\State\HubInterface::class;
     }
 }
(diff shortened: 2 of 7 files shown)


stayallive:
Hey @KarlCastle, great catch, sorry this was missed (by probably me).

Out of curiosity, why are you changing the `sentry` name. I didn't think there were any (active) packages left also using `sentry` as a container key.

But again thanks for reporting and even fixing, I've added a few more replacements, let me know if you think (or can double check) I didn't make it worse 😉

KarlCastle:
I'm supporting a legacy application that still relies on `cartalyst/sentry`. So I'm not aware of any _active_ packages left either :)

Your changes seem great. :+1:

stayallive:
Haha, sounds about right :) Well in that case thanks for taking the time to report and fix anyway 👍

stayallive (approved):

PR #474, opened 2021-04-18
Ensure user integrations are always executed after SDK integrations

baijunyao:
https://github.com/getsentry/sentry-php/pull/1210


stayallive (approved):
Thanks, I've made a little changes to simplify the code and make sure all user integrations run after the default integrations.

PR #475, opened 2021-04-22
Fix repeated booted callback registration

### LINKED ISSUE #472: Memory leak in long running process due to repeated booted callback registration
### Environment
Using  Sentry SaaS

Which SDK and version?
sentry/sentry-laravel: 2.4.2
sentry/sdk: 3.1.0
sentry/sentry: 3.2.0

### Steps to Reproduce
Running Laravel with `sentry-laravel` in long-running web worker processes such as PSR-7 based web server or Swoole.

### Expected Result
The system is stable and does not continuously increase its memory usage.

### Actual Result
After a couple of thousands requests workers start to hit the `memory_limit` of 128M.

### Identified source of problem
After about a week of debugging and narrowing it down we have identified that the issue comes from [this block of code](https://github.com/getsentry/sentry-laravel/blob/master/src/Sentry/Laravel/Tracing/Middleware.php#L103-L116), i.e. on every request the middleware would register a new booted callback until eventually exhaust all available memory.

Object tree dump after pushing 6000+ requests using `ab` (Apache Bench) tool:
```
(object) Illuminate\Foundation\Application
  basePath => (scalar) <REDACTED>
  hasBeenBootstrapped => (scalar) 1
  booted => (scalar) 1
  bootingCallbacks => (array) 1
    0 => (closure)
  bootedCallbacks => (array) 6538
    0 => (closure)
    1 => (closure)
    2 => (closure)
    3 => (closure)
    4 => (closure)
    ...
    6533 => (closure)
    6534 => (closure)
    6535 => (closure)
    6536 => (closure)
    6537 => (closure)
  terminatingCallbacks => (array) 0
  serviceProviders => (array) 39
    0 => (object) Illuminate\Events\EventServiceProvider
      app => (object) *currentApp
      bootingCallbacks => (array) 0
      bootedCallbacks => (array) 0
    ...
```

filakhtov:
Fixes #472 

The `Middleware` class was relying on booted callback to measure the amount of time it takes for application to bootstrap, observing the difference between the Laravel start time (using the defined constant) or request start time from SAPI (if one is available) and falling back onto the timestamp when the request hit middleware. There were two issues with this approach:

- once request hits middleware, the framework has already finished booting by this point in time and the registered booted callback is fired immediately anyway, so there is no reason to have a step of indirection like that;
- but secondly, and more importantly, adding booted callback on each request meant that if the middleware to be used in long running process, such as when running from RoadRunner, Swoole or PSR-7 based web server and reusing application instances, they will accumulate and eventually tripping the memory limit or be OOM-killed;

therefore, after a brief conversation in #472, we agreed on inlining the booted callback into the middleware itself, which this commit is addressing.

### DIFF
--- a/src/Sentry/Laravel/Tracing/Middleware.php
+++ b/src/Sentry/Laravel/Tracing/Middleware.php
@@ -79,7 +79,7 @@ class Middleware
 
     private function startTransaction(Request $request, HubInterface $sentry): void
     {
-        $fallbackTime = microtime(true);
+        $requestStartTime = $request->server('REQUEST_TIME_FLOAT', microtime(true));
         $sentryTraceHeader = $request->header('sentry-trace');
 
         $context = $sentryTraceHeader
@@ -91,7 +91,7 @@ class Middleware
             'url' => '/' . ltrim($request->path(), '/'),
             'method' => strtoupper($request->method()),
         ]);
-        $context->setStartTimestamp($request->server('REQUEST_TIME_FLOAT', $fallbackTime));
+        $context->setStartTimestamp($requestStartTime);
 
         $this->transaction = $sentry->startTransaction($context);
 
@@ -99,21 +99,21 @@ class Middleware
         SentrySdk::getCurrentHub()->setSpan($this->transaction);
 
         if (!$this->addBootTimeSpans() && app() instanceof Laravel) {
-            // @TODO: We might want to move this together with the `RouteMatches` listener to some central place and or do this from the `EventHandler`
-            app()->booted(function () use ($request, $fallbackTime): void {
+            $laravelStartTime = defined('LARAVEL_START') ? LARAVEL_START : $request->server('REQUEST_TIME_FLOAT');
+            if ($laravelStartTime !== null) {
                 $spanContextStart = new SpanContext();
                 $spanContextStart->setOp('app.bootstrap');
-                $spanContextStart->setStartTimestamp(defined('LARAVEL_START') ? LARAVEL_START : $request->server('REQUEST_TIME_FLOAT', $fallbackTime));
+                $spanContextStart->setStartTimestamp($laravelStartTime);
                 $spanContextStart->setEndTimestamp(microtime(true));
                 $this->transaction->startChild($spanContextStart);
+            }
 
-                $appContextStart = new SpanContext();
-                $appContextStart->setOp('app.handle');
-                $appContextStart->setStartTimestamp(microtime(true));
-                $this->appSpan = $this->transaction->startChild($appContextStart);
+            $appContextStart = new SpanContext();
+            $appContextStart->setOp('app.handle');
+            $appContextStart->setStartTimestamp(microtime(true));
+            $this->appSpan = $this->transaction->startChild($appContextStart);
 
-                SentrySdk::getCurrentHub()->setSpan($this->appSpan);
-            });
+            SentrySdk::getCurrentHub()->setSpan($this->appSpan);
         }
     }
 
--- a/src/Sentry/Laravel/Tracing/Middleware.php
+++ b/src/Sentry/Laravel/Tracing/Middleware.php
@@ -38,6 +38,8 @@ class Middleware
 
     /**
      * Set the timestamp of application bootstrap completion.
+     * For Lumen this method should be manually invoked at the end of application bootstrap process to include the
+     * `app.bootstrap` span.
      *
      * @internal This method should only be invoked from the "booted" callback
      * @param int $timestamp
@@ -117,15 +119,18 @@ class Middleware
         // Setting the Transaction on the Hub
         SentrySdk::getCurrentHub()->setSpan($this->transaction);
 
-        if (!$this->addBootTimeSpans() && app() instanceof Laravel) {
+        if (!$this->addBootTimeSpans()) {
             $this->addAppBootstrapSpan($request);
 
             $appContextStart = new SpanContext();
             $appContextStart->setOp('app.handle');
-            $appContextStart->setStartTimestamp(microtime(true));
+            $appContextStart->setStartTimestamp($this->bootedTimestamp ?? microtime(true));
             $this->appSpan = $this->transaction->startChild($appContextStart);
 
             SentrySdk::getCurrentHub()->setSpan($this->appSpan);
+
+            // Reset booted timestamp, because we don't want to report bootstrap time more than once
+            $this->bootedTimestamp = null;
         }
     }
 
@@ -146,9 +151,6 @@ class Middleware
         $spanContextStart->setStartTimestamp($laravelStartTime);
         $spanContextStart->setEndTimestamp($this->bootedTimestamp);
         $this->transaction->startChild($spanContextStart);
-
-        // Reset booted timestamp, because we don't want to report bootstrap time more than once
-        $this->bootedTimestamp = null;
     }
 
     private function addBootTimeSpans(): bool
--- a/src/Sentry/Laravel/Tracing/Middleware.php
+++ b/src/Sentry/Laravel/Tracing/Middleware.php
@@ -120,46 +120,15 @@ class Middleware
         // Setting the Transaction on the Hub
         SentrySdk::getCurrentHub()->setSpan($this->transaction);
 
-        if (!$this->addBootTimeSpans()) {
-            $bootstrapSpan = $this->addAppBootstrapSpan($request);
+        $bootstrapSpan = $this->addAppBootstrapSpan($request);
 
-            $appContextStart = new SpanContext();
-            $appContextStart->setOp('app.handle');
-            $appContextStart->setStartTimestamp($bootstrapSpan ? $bootstrapSpan->getEndTimestamp() : microtime(true));
+        $appContextStart = new SpanContext();
+        $appContextStart->setOp('app.handle');
+        $appContextStart->setStartTimestamp($bootstrapSpan ? $bootstrapSpan->getEndTimestamp() : microtime(true));
 
-            $this->appSpan = $this->transaction->startChild($appContextStart);
+        $this->appSpan = $this->transaction->startChild($appContextStart);
 
-            SentrySdk::getCurrentHub()->setSpan($this->appSpan);
-        }
-    }
-
-    private function addBootTimeSpans(): bool
-    {
-        if (!defined('LARAVEL_START') || !LARAVEL_START) {
-            return false;
-        }
-
-        if (!defined('SENTRY_AUTOLOAD') || !SENTRY_AUTOLOAD) {
-            return false;
-        }
-
-        if (!defined('SENTRY_BOOTSTRAP') || !SENTRY_BOOTSTRAP) {
-            return false;
-        }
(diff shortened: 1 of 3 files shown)


src/Sentry/Laravel/Tracing/Middleware.php
```diff
@@ -91,29 +91,29 @@ private function startTransaction(Request $request, HubInterface $sentry): void
...
-            app()->booted(function () use ($request, $fallbackTime): void {
+            $laravelStartTime = defined('LARAVEL_START') ? LARAVEL_START : $request->server('REQUEST_TIME_FLOAT');
+            if ($laravelStartTime !== null) {
                 $spanContextStart = new SpanContext();
                 $spanContextStart->setOp('app.bootstrap');
-                $spanContextStart->setStartTimestamp(defined('LARAVEL_START') ? LARAVEL_START : $request->server('REQUEST_TIME_FLOAT', $fallbackTime));
+                $spanContextStart->setStartTimestamp($laravelStartTime);
                 $spanContextStart->setEndTimestamp(microtime(true));
                 $this->transaction->startChild($spanContextStart);
+            }
```
filakhtov, line 109: Note, this is a change compared to the initial implementation and I am open to be challenged on this assumption here, but, if neither `LARAVEL_START` constant is defined nor `REQUEST_TIME_FLOAT` SAPI is available, we should not just use timestamp of the start of this middleware, because the difference between these two values will be negligible and will misreport the "bootstrap" and will be misleading for those who observing timings, making them think that bootstrap was actually nearly instantaneous, while that is not even remotely true.

src/Sentry/Laravel/Tracing/Middleware.php
```diff
@@ -79,7 +79,7 @@ public function terminate($request, $response): void
 
     private function startTransaction(Request $request, HubInterface $sentry): void
     {
-        $fallbackTime = microtime(true);
+        $requestStartTime = $request->server('REQUEST_TIME_FLOAT', microtime(true));
```
stayallive, line 82: Not every server has the request time set, so we should fallback to `LARAVEL_START` and only the use the time we entered the middleware as the last resort.

`LARAVEL_START` should be a whole lot closer to the actual start time of the request since it's at the top of the `index.php` before the autoloader when we arrive here in `startTransaction` there was already quite some work done.

```suggestion
        $requestStartTime = $request->server('REQUEST_TIME_FLOAT', defined('LARAVEL_START') ? LARAVEL_START : microtime(true));
```

src/Sentry/Laravel/Tracing/Middleware.php
stayallive: Thinking about this some more and you are right, however I feel like the bootstrap section is not being made longer then it needs to be too since ending up in `startTransaction` is not the actual `booted` time of the app but a little after it.

Since the booted event happens before the transaction is even started we would have to buffer it. I might be going to say some dirty dirty words here but what if we would add a booted event callback in the event listener and set a static property on the middleware to have the actual timestamp of when the booted event occurred we can use here to set the actual end timestamp, the start timestamp is fine like how you did it I feel.

We can also then set the start for the `app.handle` frame to the end timestamp of the bootstrap section since they come right after each other.

src/Sentry/Laravel/Tracing/Middleware.php
filakhtov: Unless I am missing something it was this way all along: https://github.com/getsentry/sentry-laravel/pull/475/files#diff-fd513f75c5631d247590ec3dc4e2df94b986aab48d02af48958a6aec8e0f176fL94-R94 and for bootstrap time we got https://github.com/getsentry/sentry-laravel/pull/475/files#diff-fd513f75c5631d247590ec3dc4e2df94b986aab48d02af48958a6aec8e0f176fR102

stayallive: Yeah but while we are at it I figured we might as well improve it good right?

If we don't have the `REQUEST_TIME_FLOAT` a more sensible time would be `LARAVEL_START` instead of immeditely resort to `microtime()` wouldn't you agree?

src/Sentry/Laravel/Tracing/Middleware.php
filakhtov: The static property would actually work... but we can also have a shared stateful dependency that's injected into the middleware and used by the callback to achieve this same result without it being too "dirty". Happy to do that if you also think that is by any means better.

As to:
> We can also then set the start for the `app.handle` frame to the end timestamp of the bootstrap section since they come right after each other.

This would actually break things for long running processes, because we bootstrap once and then handle request from then on, and I think it is better to have transaction start time originating from middleware. What do you say?

src/Sentry/Laravel/Tracing/Middleware.php
filakhtov: I am absolutely on board with that, just wanted to make sure that was an original oversight, not something broken or omitted by myself. I will push an amendment for that!

filakhtov: Actually, thinking on this further, this be a problem for long running processes, because `LARAVEL_START` will be defined at the worker start, which will be far away from the request start...

stayallive: Good point, in that case I hope they all set the `REQUEST_TIME_FLOAT`, if not we might need to add some more specific checks and fallbacks here for things like Octane/Swoole and others.

filakhtov:
@stayallive, thanks a lot for all the guidance 🙇🏻‍♂️ I have just pushed another commit, this time relying on a separate "booted" callback that is registered within a service provider and sets the state on Middleware to consume when appropriate.

src/Sentry/Laravel/Tracing/ServiceProvider.php
```diff
@@ -36,6 +36,10 @@ public function boot(): void
     public function register(): void
     {
         $this->app->singleton(Middleware::class);
+
+        $this->app->booted(function () {
+            $this->app->make(Middleware::class)->setBootedTimestamp(microtime(true));
+        });
```
stayallive, line 42: So this won't work with Lumen since it doesn't have a `booted` callback...

I guess we could do something like this for Lumen and just accept there is no booting span.

```suggestion
		if (!$this->app instanceof Lumen) {
	        $this->app->booted(function () {
    	        $this->app->make(Middleware::class)->setBootedTimestamp(microtime(true));
        	});
        }
```

src/Sentry/Laravel/Tracing/Middleware.php
```diff
@@ -91,32 +110,47 @@ private function startTransaction(Request $request, HubInterface $sentry): void
...
-                $appContextStart->setStartTimestamp(microtime(true));
-                $this->appSpan = $this->transaction->startChild($appContextStart);
-
-                SentrySdk::getCurrentHub()->setSpan($this->appSpan);
-            });
+            $this->addAppBootstrapSpan($request);
+
+            $appContextStart = new SpanContext();
+            $appContextStart->setOp('app.handle');
+            $appContextStart->setStartTimestamp(microtime(true));
```
stayallive, line 125: Okay, how about this, because you make some good points but I feel `microtime(true)` here is still too slow and can be better.

So what if we return the `Span` from the `addAppBootstrapSpan` and IF it's returned (happens because if we don't do the bootstrap span because no booted timestamp or some other reason) we default to `microtime(true)`?

Something like this:

```suggestion
            $appBootedSpan = $this->addAppBootstrapSpan($request);

            $appContextStart = new SpanContext();
            $appContextStart->setOp('app.handle');
            $appContextStart->setStartTimestamp($appBootedSpan ? $appBootedSpan->getStartTimestamp() : microtime(true));
```

And change `addAppBootstrapSpan` to return `Span|null` of course?

stayallive (commented):
Sorry for the massive nitpicking on just 50 lines of code but you made me realize there is a lot left on the table so I hope you don't mind too much 😉

src/Sentry/Laravel/Tracing/ServiceProvider.php
filakhtov: This is a very good point. I did not think about `Lumen` - we had this check in `Middleware` and I should have moved it into the service provider. With this change we also make it _easier_ for Lumen to enable this metric - they just can invoke `setBootedTimestamp()` from whatever routine is appropriate and `Middleware` will do the rest.

src/Sentry/Laravel/Tracing/Middleware.php
filakhtov: 🤔 this is where I have a question: does `app.handle` includes `app.bootstrap` or should `app.handle` start _after_ the `app.bootstrap`? Because according to your suggestion we would set start timestamp for both `app.handle` and `app.bootstrap` to the same value and _I think_ we actually should set `app.handle` start time to `app.bootstrap` end time - at least based on the original logic that we had in place before I started messing around with it? What do you think?

filakhtov:
@stayallive I have pushed another commit with two changes:
- fixing support for `Lumen` by adding the check around the \`booted()\` callback
- setting `app.handle` start timestamp to `app.bootstrap` end timestamp, but this is an assumption that I would like you to verify and ensure that's what we actually want, instead of them both being set on the same start time.

src/Sentry/Laravel/Tracing/Middleware.php
```diff
@@ -29,6 +29,27 @@ class Middleware
...
+     * The timestamp of application bootstrap completion.
+     *
+     * @var null|int
+     */
+    private $bootedTimestamp = null;
+
+    /**
+     * Set the timestamp of application bootstrap completion.
+     * For Lumen this method should be manually invoked at the end of application bootstrap process to include the
+     * `app.bootstrap` span.
```
filakhtov, line 42: Also added this note for \`Lumen\` users.

src/Sentry/Laravel/Tracing/Middleware.php
stayallive: Yes that was what I meant to say 👍  (typo on the start time)

stayallive:
I made some final changes and clean it a bit up, but this looks very good to me!

image

Also looks like it still does what it's supposed to do 👍 

Thanks for taking the time and effort the create and improve this PR, much appreciated.

stayallive (approved):
Please check it out if my changes are still fine (I think they are) and with your approval this is ready for release!

stayallive:
I actually made a few more changes which would result in this when adding the `SENTRY_AUTOLOAD` define. The bootstrap define did not make any sense to me so I've removed it.

image

filakhtov:
👏🏻 oh nice, thank you very much for that autoload refactoring - it makes much more sense now! Please, go ahead and merge this at your convenience.

A question from me: what is the release cycle looks like, i.e. how long before this change would actually be released under a version tag?

stayallive:
There is not a set release cycle but today or tomorrow is when I want to tag a new release so it should not be long (but no promises just in case 😁).

filakhtov:
Thanks, and likewise, appreciate your support and guidance on the way there! I will stay tuned for the release then! 🧐

PR #484, opened 2021-04-29
Replace usage of removed Sentry\FlushableClientInterface with Sentry\ClientInterface

superbiche:
This fixes an issue where the condition `if ($client instanceof FlushableClientInterface) {` always returns `false`, as Sentry SDK removed `FlushableClientInterface` in 3.0.0 - see https://github.com/getsentry/sentry-php/blob/master/CHANGELOG.md#300-2020-09-28

Replacing with `ClientInterface` which is implemented by `Client` and also enforces the `flush` method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Correctly call flush on the PHP SDK client (#484)
+
 ## 2.5.1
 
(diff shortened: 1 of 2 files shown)


stayallive (approved):
Good catch, although with the current "stock" transport this does nothing because it sends immediately, for custom transports this could still be important 👍

PR #490, opened 2021-05-18
Laravel Lighthouse tracing integration

stayallive:
This PR aims to add out of the box tracing support for Laravel Lighthouse.

Sneak peek of how a batched request executing multiple GraphQL queries would look like:

image

/cc @spawnia

### DIFF
--- a/src/Sentry/Laravel/ServiceProvider.php
+++ b/src/Sentry/Laravel/ServiceProvider.php
@@ -7,6 +7,7 @@ use Illuminate\Foundation\Application as Laravel;
 use Illuminate\Foundation\Http\Kernel as HttpKernel;
 use Illuminate\Log\LogManager;
 use Laravel\Lumen\Application as Lumen;
+use RuntimeException;
 use Sentry\ClientBuilder;
 use Sentry\ClientBuilderInterface;
 use Sentry\Integration as SdkIntegration;
@@ -23,9 +24,10 @@ class ServiceProvider extends BaseServiceProvider
      * List of configuration options that are Laravel specific and should not be sent to the base PHP SDK.
      */
     private const LARAVEL_SPECIFIC_OPTIONS = [
-        // We do not want these settings to hit the PHP SDK because it's Laravel specific and the SDK will throw errors
+        // We do not want these settings to hit the PHP SDK because they are Laravel specific and the PHP SDK will throw errors
         'tracing',
         'breadcrumbs',
+        'tracing_integrations',
         // We resolve the integrations through the container later, so we initially do not pass it to the SDK yet
         'integrations',
         // This is kept for backwards compatibility and can be dropped in a future breaking release
@@ -132,7 +134,7 @@ class ServiceProvider extends BaseServiceProvider
         }
 
         $this->app->bind(ClientBuilderInterface::class, function () {
-            $basePath = base_path();
+            $basePath   = base_path();
             $userConfig = $this->getUserConfig();
 
             foreach (self::LARAVEL_SPECIFIC_OPTIONS as $laravelSpecificOptionName) {
@@ -141,7 +143,7 @@ class ServiceProvider extends BaseServiceProvider
 
             $options = \array_merge(
                 [
-                    'prefixes' => [$basePath],
+                    'prefixes'       => [$basePath],
                     'in_app_exclude' => ["{$basePath}/vendor"],
                 ],
                 $userConfig
@@ -233,16 +235,27 @@ class ServiceProvider extends BaseServiceProvider
         foreach ($userIntegrations as $userIntegration) {
             if ($userIntegration instanceof SdkIntegration\IntegrationInterface) {
                 $integrations[] = $userIntegration;
-            } elseif (\is_string($userIntegration)) {
+            } elseif (\is_string($userIntegration) && $this->app->bound($userIntegration)) {
                 $resolvedIntegration = $this->app->make($userIntegration);
 
-                if (!($resolvedIntegration instanceof SdkIntegration\IntegrationInterface)) {
-                    throw new \RuntimeException('Sentry integrations should a instance of `\Sentry\Integration\IntegrationInterface`.');
+                if (!$resolvedIntegration instanceof SdkIntegration\IntegrationInterface) {
+                    throw new RuntimeException(
+                        sprintf(
+                            'Sentry integrations must be an instance of `%s` got `%s`.',
+                            SdkIntegration\IntegrationInterface::class,
+                            get_class($resolvedIntegration)
+                        )
+                    );
                 }
 
                 $integrations[] = $resolvedIntegration;
             } else {
-                throw new \RuntimeException('Sentry integrations should either be a container reference or a instance of `\Sentry\Integration\IntegrationInterface`.');
+                throw new RuntimeException(
+                    sprintf(
+                        'Sentry integrations must either be a valid container reference or an instance of `%s`.',
+                        SdkIntegration\IntegrationInterface::class
+                    )
+                );
             }
         }
 
--- a/src/Sentry/Laravel/ServiceProvider.php
+++ b/src/Sentry/Laravel/ServiceProvider.php
@@ -28,7 +28,6 @@ class ServiceProvider extends BaseServiceProvider
         // We do not want these settings to hit the PHP SDK because they are Laravel specific and the PHP SDK will throw errors
         'tracing',
         'breadcrumbs',
-        'tracing_integrations',
         // We resolve the integrations through the container later, so we initially do not pass it to the SDK yet
         'integrations',
         // This is kept for backwards compatibility and can be dropped in a future breaking release
--- a/src/Sentry/Laravel/ServiceProvider.php
+++ b/src/Sentry/Laravel/ServiceProvider.php
@@ -14,6 +14,7 @@ use Sentry\Integration as SdkIntegration;
 use Sentry\Laravel\Http\LaravelRequestFetcher;
 use Sentry\Laravel\Http\SetRequestIpMiddleware;
 use Sentry\Laravel\Http\SetRequestMiddleware;
+use Sentry\Laravel\Tracing\ServiceProvider as TracingServiceProvider;
 use Sentry\SentrySdk;
 use Sentry\State\Hub;
 use Sentry\State\HubInterface;
@@ -230,9 +231,17 @@ class ServiceProvider extends BaseServiceProvider
             new Integration\ExceptionContextIntegration,
         ];
 
-        $userIntegrations = $this->getUserConfig()['integrations'] ?? [];
+        $userConfig = $this->getUserConfig();
 
-        foreach ($userIntegrations as $userIntegration) {
+        $integrationsToResolve = $userConfig['integrations'] ?? [];
+
+        $enableDefaultTracingIntegrations = $userConfig['tracing']['default_integrations'] ?? true;
+
+        if ($enableDefaultTracingIntegrations && $this->couldHavePerformanceTracingEnabled()) {
+            $integrationsToResolve = array_merge($integrationsToResolve, TracingServiceProvider::DEFAULT_INTEGRATIONS);
+        }
+
+        foreach ($integrationsToResolve as $userIntegration) {
             if ($userIntegration instanceof SdkIntegration\IntegrationInterface) {
                 $integrations[] = $userIntegration;
             } elseif (\is_string($userIntegration)) {
--- a/src/Sentry/Laravel/ServiceProvider.php
+++ b/src/Sentry/Laravel/ServiceProvider.php
@@ -235,7 +235,7 @@ class ServiceProvider extends BaseServiceProvider
         foreach ($userIntegrations as $userIntegration) {
             if ($userIntegration instanceof SdkIntegration\IntegrationInterface) {
                 $integrations[] = $userIntegration;
-            } elseif (\is_string($userIntegration) && $this->app->bound($userIntegration)) {
+            } elseif (\is_string($userIntegration)) {
                 $resolvedIntegration = $this->app->make($userIntegration);
 
                 if (!$resolvedIntegration instanceof SdkIntegration\IntegrationInterface) {
--- a/src/Sentry/Laravel/Tracing/Integrations/IntegrationInterface.php
+++ b/src/Sentry/Laravel/Tracing/Integrations/IntegrationInterface.php
@@ -4,5 +4,5 @@ namespace Sentry\Laravel\Tracing\Integrations;
 
 interface IntegrationInterface
 {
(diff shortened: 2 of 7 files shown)


src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
```diff
@@ -0,0 +1,174 @@
...
+
+        ksort($groupedOperations, SORT_STRING);
+
+        $transactionName = 'lighthouse?' . implode('&', $groupedOperations);
+
+        Integration::setTransaction($transactionName);
+        SentrySdk::getCurrentHub()->getTransaction()->setName($transactionName);
+    }
+
+    private function extractOperationNames(OperationDefinitionNode $operation): array
```
spawnia, line 150:
```suggestion
    /**
     * @return array<int, string>
     */
    private function extractTopLevelFieldNames(OperationDefinitionNode $operation): array
```

src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
```diff
@@ -0,0 +1,174 @@
...
+    {
+        if ($operation->name !== null) {
+            return [$operation->name->value];
+        }
+
+        $selectionSet = [];
+
+        /** @var \GraphQL\Language\AST\FieldNode $selection */
+        foreach ($operation->selectionSet->selections as $selection) {
+            $selectionSet[] = $selection->alias === null
```
spawnia, line 160:
```suggestion
            // Not respecting aliases because they are only relevant for clients
            // and the tracing we extract here is targeted at server developers.
            $selectionSet[] = $selection->name->value;
```

src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
```diff
@@ -0,0 +1,174 @@
...
+
+    /** @var \Sentry\Tracing\Span|null $requestSpan */
+    private $requestSpan;
+
+    /** @var \Sentry\Tracing\Span|null $operationSpan */
+    private $operationSpan;
+
+    public function __construct(Dispatcher $evenDispatcher)
+    {
+        if (!class_exists(StartRequest::class)) {
```
spawnia, line 31: Would it make sense to do this detection before even instantiating the Integration? For coherence, the main service provider could call a static method within this class.

src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
```diff
@@ -0,0 +1,174 @@
...
+            sort($operations, SORT_STRING);
+
+            $operations = "{$operationType}{" . implode(',', $operations) . '}';
+        });
+
+        ksort($groupedOperations, SORT_STRING);
+
+        $transactionName = 'lighthouse?' . implode('&', $groupedOperations);
+
+        Integration::setTransaction($transactionName);
```
spawnia, line 146: So this replaces the HTTP transaction?

src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
```diff
@@ -0,0 +1,174 @@
...
+    /** @var \Sentry\Tracing\Span|null $previousSpan */
+    private $previousSpan;
+
+    /** @var \Sentry\Tracing\Span|null $requestSpan */
+    private $requestSpan;
+
+    /** @var \Sentry\Tracing\Span|null $operationSpan */
+    private $operationSpan;
+
+    public function __construct(Dispatcher $evenDispatcher)
```
spawnia, line 29:
```suggestion
    public function __construct(EventDispatcher $eventDispatcher)
```

I tend to prefer aliases for those super generic Laravel class names.

spawnia:
I think it would be better to use only the top-level fields for naming and grouping operations. That makes it very predictable, whereas using operation names and aliases is dependent on the client and might change on a whim.

If tracing on the level of named operations is wanted, this can still happen from the client side. In integrated scenarios, the clients are most likely using Sentry too 😉

src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
stayallive: I actually had that, and that might still be the better option. I'm not a 100% sure, this class exists feels a little too brittle too for my liking... some version check would be a lot better IMO but I'm not sure we can do that without adding dependencies or parsing the composer.json... still thinking about this 🧐

src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
stayallive: Yes. The reason is not too stupid but this is not working as intended at the moment. 

Traces and actual errors are matched based on the transaction name. But since we only know the new transaction name after errors happened this is not doing what I want it to do at the moment although I also don't think this is achievable, needs a little more research to see what the drawbacks of non-matching transaction names are like they would be in this case.

stayallive:
Yeah, I do not agree on that. The operation name is super useful IMHO, but I also see where you are coming from and I might even consider adding a configuration option or something else so you can switch between the two without too much work on the integrator.

src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
spawnia: I was thinking if we should add a canonical way of detecting if Lighthouse is installed to Lighthouse itself. On the other hand, this duck typing approach works nicely as it tests for the exact feature that is required.

src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
spawnia: What is stopping us from setting the transaction name right when the request starts?

spawnia:
I can see how both approaches might be desirable in different scenarios. We have multiple clients using our API, so using operation names does not work well for us. It might group fields that are completely distinct but coincidentally named the same, and fail to group fields that are actually the same but named differently.

src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
stayallive: Also true, although it won't account for if in a major version of Ligthouse the even gains or loses properties we depend on, so we either need to program very defensively for that (this integration should _never_ crash) or we need to make sure we are on the correct version.

This is of course not limited to Ligthouse but to any integration we might add in the future.

src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
stayallive: `StartRequest` does not contain the operations we would need to do that and `StartExecution` is for a single execution so a batched request is not considered there yet. At this moment I believe we only have all operations at the end of the last `EndExecution` which we don't know it's the last one so we can only be sure we seen it all after `EndRequest`. But maybe I'm missing something?

stayallive:
Yeah exactly, I think having both is the best option here because it would depend on the use cases. For us the difference between `query{viewer}` and `query{viewer}` would be desirable and could be indicated by the operation or query name because although they only query the `viewer` they can have a vastly different subselection of fields.

src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
spawnia: Got it, makes sense that we only have all the information after parsing all requests.

src/Sentry/Laravel/Tracing/ServiceProvider.php
```diff
@@ -115,4 +119,37 @@ private function wrapViewEngine(Engine $realEngine): Engine
...
+        $integrations = array_merge(
+            $defaultIntegrations,
+            $this->getUserConfig()['tracing_integrations'] ?? []
+        );
+
+        /** @var \Sentry\Laravel\Tracing\Integrations\IntegrationInterface $tracingIntegration */
+        foreach ($integrations as $tracingIntegration) {
+            if (!is_subclass_of($tracingIntegration, Integrations\IntegrationInterface::class)) {
+                throw new RuntimeException(
+                    sprintf('Sentry tracing integrations should an instance of `%s`.', Integrations\IntegrationInterface::class)
```
spawnia, line 140:
```suggestion
                    sprintf('Sentry tracing integrations must be an instance of `%s`.', Integrations\IntegrationInterface::class)
```

src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
stayallive: Me too, I don't think that is the right idea either. I will not touch the request transaction name for now and just change the transaction transaction name (yes I know confusing 😉).

src/Sentry/Laravel/Tracing/Integrations/IntegrationInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Sentry\Laravel\Tracing\Integrations;
+
+interface IntegrationInterface
+{
+    public static function supported(): bool;
```
spawnia, line 7: Not sure if the notion of being supported is an exact semantic fit here, perhaps something like `isApplicable()` would be more expressive?

spawnia:
I am quite excited to try this out!

src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
spawnia: To ask in terms I understand: when clicking the *Performance* tab in Sentry, will I see more then a single transaction `graphql`?

image

stayallive: Yes.

When you click into the transaction there is a `Related Issues` block. That only works if the transaction name (in your screenshot `graphql`) matches up between the transaction and the error.

Because we modify the transaction name (for the transactions not the errors/logs anymore) we break that link, it results in multiple transactions (grouped on transaction name) which is 100% what we want but related errors will be empty unless we figure out a good way to know the transaction name earlier so error/log events will have the correct transaction name set.

Does this make sense?

stayallive:
So what about this for changing if the integration ignores operation names?

```diff
<?php

namespace App\Providers;

use Illuminate\Support\ServiceProvider;

class AppServiceProvider extends ServiceProvider
{
    public function register()
    {
+       $this->app->when(
+           \Sentry\Laravel\Tracing\Integrations\LighthouseIntegration::class
+       )->needs('$ignoreOperationName')->give(true);
    }
}
```

This seems like a pretty cool solution to prevent needing to make life difficult?

src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
spawnia: I guess `StartOperationOrOperations` is starting to look more and more juicy 😉

stayallive: I must admit it doesn't sound too bad anymore 😅

Been thinking on other ways. But there is no other way to do it unless we independently parse the GraphQL request on StartRequest which is a terrible idea. 

I'll leave it up to you to decide on it 😬

spawnia:
@stayallive Woah, that usage of the Container is pretty advanced. Looks acceptable to me.

stayallive:
I have started testing this branch on our staging environment and I'm pretty excited for this already!

image

You can test it by making the following change in your `composer.json` (and reverting that later!):

```diff
-   "sentry/sentry-laravel": "^2.5",
+   "sentry/sentry-laravel": "dev-lighthouse-integration#6cb6bfab3d03869db80022bb9a4c3c766dd83fca",
```

No other configuration should be needed "out of the box" apart from setting up tracing in the first place.

If you want to ignore operation names passed by the client and want to group transactions on the root selection set instead you can add the following snippet to the `register` method of for example your `AppServiceProvider`:

```php
$this->app->when(
    \Sentry\Laravel\Tracing\Integrations\LighthouseIntegration::class
)->needs('$ignoreOperationName')->give(true);
```

src/Sentry/Laravel/ServiceProvider.php
```diff
@@ -23,9 +24,10 @@ class ServiceProvider extends BaseServiceProvider
      * List of configuration options that are Laravel specific and should not be sent to the base PHP SDK.
      */
     private const LARAVEL_SPECIFIC_OPTIONS = [
-        // We do not want these settings to hit the PHP SDK because it's Laravel specific and the SDK will throw errors
+        // We do not want these settings to hit the PHP SDK because they are Laravel specific and the PHP SDK will throw errors
         'tracing',
         'breadcrumbs',
+        'tracing_integrations',
```
HazAT, line 30: Is there a reason why we want to separate this into tracing integrations? I would prefer to have it just as an integration to not intro something new.

stayallive: Good question actually... this was mostly to be able to keep it separate, all tracing code is registered using a second service provider in Laravel. If we add the tracing integrations as SDK integrations that would mean it needs to be added to the non-tracing service provider. 

But technically there is nothing stopping us from using the normal integrations. So we should probably do that!

HazAT: I would prefer to only have `intergrations` and enable this depending on `traces_sample_rate` option.

stayallive: Refactored it to do that 👍

src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
stayallive: So... after thinking a bit more about this I settled on: f9eebd3be86a0158dc9ad87623c54fa26a5bf23b.

This might cause incorrect transaction names on batch requests but for all other instances this should work.

No need for `StartOperationOrOperations` unless batch request transaction names also need to be correct, so it would still be an improvement to have something like that.

spawnia:
https://github.com/nuwave/lighthouse/releases/tag/v5.11.0 is ready.

src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
spawnia: https://github.com/nuwave/lighthouse/releases/tag/v5.11.0 is ready.

spawnia:
Before and after in this screenshot, it is a great improvement:

image

spawnia (approved):
Eagerly awaiting the release 🚢

PR #496, opened 2021-06-10
SentryHandler: Replacing concrete type Hub with abstract HubInterface

mingalevme:
Currently `SentryHandler` uses concrete (and also **final**) class `Hub`, but it should use corresponding interface `HubInterface`.


stayallive:
The change makes perfect sense, good change.

But the test... I'm not sure what is event tested it feels like we are testing basic PHP type hints here with a lot of code. Can we do without the test?

mingalevme:
@stayallive 

Yes, test is ugly, but I have not come up with anything better.

So If you can merge without the test and no new pull request is required - you're welcome.

stayallive (approved):
Thanks for this, good change 👍

PR #499, opened 2021-06-16
Update to use sentry/sentry 3.3

3Dgoo:


stayallive (approved):
Thanks, that was missed 👍

stayallive:
I've also relaxed the constraint further to just `^3.3` for now to prevent these issues in the future.

PR #504, opened 2021-07-09
Add Tracing capabilities to Facade

pataar:
These methods are available on the HubInterface, therefore they can be added to this Facade.


stayallive (approved):
Thanks, I've re-ordered them and fixed some outdated definitions too to match the `HubInterface` class exactly.

PR #509, opened 2021-08-02
Add options to the `artisan sentry:publish` command to better support `--no-interaction` mode

vokamut:


vokamut:
Because the default values will change

stayallive (approved):
Yep you are right, better to keept the same defaults :)

Thanks for this, I only changes to name of the performance monitoring option to be more clear IMHO.

Issue #514, opened 2021-09-28
Checking offset "level" on null value


wimulkeman:
### Environment

How do you use Sentry?
Sentry SAAS

Which SDK and version?
php - "sentry/sentry-laravel": "^2.8",

### Steps to Reproduce

1. Wrap the Sentry Handler in FingersCrossedHandler
2. Log errors below the activation level from the FingersCrossedHandler
3. Log error above the activation level from the FingersCrossedHandler

### Expected Result

I tried to log one entry with a batch of debug log entries in Sentry without an error.

### Actual Result

In the code I encapsulated the SentryHandler to use the FingersCrossedHandler. This enables the option to buffer debug log messages. When an log event from the defined action_level or higher is set in the application all the previous buffered messages are flushed into the logs to provide context to the eventual error.

```php
<?php

namespace App\Log;

use Monolog\Handler\FingersCrossedHandler;
use Monolog\Handler\HandlerInterface;
use Sentry\Laravel\LogChannel as SentryLogChannel;

class SentryActionLevelLogChannel extends SentryLogChannel
{
    protected function prepareHandler(HandlerInterface $handler, array $config = [])
    {
        $handler = parent::prepareHandler($handler, $config);

        if (!empty($config['action_level'])) {
            $handler = new FingersCrossedHandler($handler, $config['action_level']);
        }

        return $handler;
    }
}
```

When this flush is triggered, the following method is called: [SentryHandler::handleBatch](https://github.com/getsentry/sentry-laravel/blob/master/src/Sentry/Laravel/SentryHandler.php#L73)

This loops through the errors and logs the log entry with the highest error level in Sentry. The other log entries are logged as context of the entry with the highest level.

https://github.com/getsentry/sentry-laravel/blob/master/src/Sentry/Laravel/SentryHandler.php#L90

```php
        // the record with the highest severity is the "main" one
        $record = array_reduce(
            $records,
            function ($highest, $record) {
                if ($record['level'] > $highest['level']) {
                    return $record;
                }

                return $highest;
            }
        );
```

During the first run the `$highest` variable contains `null` which throws an error when trying to fetch the index `level`.

### Possible solution

First check if the index is available.

```php
        // the record with the highest severity is the "main" one
        $record = array_reduce(
            $records,
            function ($highest, $record) {
                if (!isset($highest['level']) || $record['level'] > $highest['level']) {
                    return $record;
                }

                return $highest;
            }
        );
```

wimulkeman:
[...]

PR #515, opened 2021-09-28
fix: check index is availability in method SentryHandler::handleBatch

wimulkeman:
When flushing logrecords with the handleBatch method the code now checks if $highets['level'] is available before using it.

fixes getsentry/sentry-laravel#514

### DIFF
--- a/src/Sentry/Laravel/SentryHandler.php
+++ b/src/Sentry/Laravel/SentryHandler.php
@@ -90,7 +90,9 @@ class SentryHandler extends AbstractProcessingHandler
         $record = array_reduce(
             $records,
             function ($highest, $record) {
-                if ($record['level'] > $highest['level']) {
+                if (!isset($highest['level'])
+                    || $record['level'] > $highest['level']
+                ) {
                     return $record;
                 }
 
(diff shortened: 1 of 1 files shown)


src/Sentry/Laravel/SentryHandler.php
```diff
@@ -90,7 +90,9 @@ function ($record) use ($level) {
         $record = array_reduce(
             $records,
             function ($highest, $record) {
-                if ($record['level'] > $highest['level']) {
+                if (!isset($highest['level'])
+                    || $record['level'] > $highest['level']
+                ) {
```
stayallive, line 95: In your original report (#514) you mention `$highest` is `null` so why not check for that instead of checking if the `level` key is missing since it should never miss on a Monolog record IIRC.

```suggestion
                if ($highest === null || $record['level'] > $highest['level']) {
```

stayallive (commented):
Thanks for taking the time to create this PR!

src/Sentry/Laravel/SentryHandler.php
wimulkeman: Good point. That would make the code cleaner and signal the issue it addresses more clearly. I've changed it.

Issue #516, opened 2021-09-29
Enable log debug record concatenation 


wimulkeman:
## Feature

In Monolog it is possible to use the `FingersCrossedHandler`. When this handler is used a user can set the option `action_level` in their config.

When the `FingersCrossedHandler` is used the following feature becomes available: Monolog will buffer all logrecords which match the defined `level`, but not write them to the log file. When a log record is given which is at or above the `action_level` the buffered records will be flushed to the log.

This feature enables the user to gather debug information in their codebase and only flush it to the logs when the actual error occurs.

## Current support

Sentry already supports this up to a certain point because the `handleBatch` method is implemented in the `SentryHandler` class. When you custom wrap the `SentryHandler` in the `FingersCrossedHandler`  the functionality becomes available. At the moment this requires an extension of `Sentry\Laravel\LogChannel` like the following for example.

```php
<?php

namespace App\Log;

use Monolog\Handler\FingersCrossedHandler;
use Monolog\Handler\HandlerInterface;
use Sentry\Laravel\LogChannel as SentryLogChannel;

class SentryActionLevelLogChannel extends SentryLogChannel
{
    protected function prepareHandler(HandlerInterface $handler, array $config = [])
    {
        $handler = parent::prepareHandler($handler, $config);

        if (!empty($config['action_level'])) {
            $handler = new FingersCrossedHandler($handler, $config['action_level']);
        }

        return $handler;
    }
}
```

## Suggested support

It would be a nice improvement if a user can provide the `action_level` configuration and `sentry-laravel` will act upon it by wrapping the `SentryHandler` in a `FingersCrossedHandler`.

PR #517, opened 2021-09-29
Solves #516: Enable action_level configuration for using FingersCrossedHandler

wimulkeman:
This update enables the user to provide the Sentry log channel with an action_level configuration. The inspiration for the configuration is taken from the Symfony Monolog configuration style.

All records at or above the provided level configuration will be buffered but not written to the logs. When a log record with a level at or above the action_level configuration is provided all the previously buffered log records will be flushed as context['logs'] and under Breadcrumbs in the Sentry log record.

closes getsentry/sentry-laravel#516
See https://symfony.com/doc/current/logging.html#handlers-that-modify-log-entries

### DIFF
--- a/src/Sentry/Laravel/LogChannel.php
+++ b/src/Sentry/Laravel/LogChannel.php
@@ -2,6 +2,7 @@
 
 namespace Sentry\Laravel;
 
+use Monolog\Handler\FingersCrossedHandler;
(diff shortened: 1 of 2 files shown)


stayallive (approved):
Looks good, thanks! 👍

PR #522, opened 2021-11-10
Improve output and DX for `sentry:test` and `sentry:publish` commands

### LINKED ISSUE #521: Add a specific error message for SSL errors in the test command
Currently when we run the test command, if there are SSL errors, a [generic message](https://github.com/getsentry/sentry-laravel/blob/48909d2a174c5af4d78f8b935c071431fd38cb6f/src/Sentry/Laravel/TestCommand.php#L58) appears: `Please check if your DSN is set properly in your config or '.env' as 'SENTRY_LARAVEL_DSN'` . This does not help in pin-pointing the actual issue and it delays troubleshooting, therefore it would be useful to have a specific message for those kind of errors.

stayallive:
Things fixed / improved for `sentry:publish`:

- Ignore commented out env variables and set new ones instead of updating a commented out variable
- Set `SENTRY_TRACES_SAMPLE_RATE` to `0` if transaction should be disabled and `SENTRY_TRACES_SAMPLE_RATE` is present in the `.env`
- Handle missing DSN when command is run non-interactively
- When DSN is provided with the env key in front of it, strap it away no matter the env key (just use whatever comes after the first `=`)
- Fix checking if `.env` key is set (never worked)
- Only make changes to the `.env` after `sentry:test` is a success (or no test is performed)
- Return a unix exit code depending on if the command succeeded or failed

Things fixed / improved for `sentry:test`:

- Use a client specific for the test command so we can be a little more in control
- Log internal SDK errors and show them as errors when sending fails
- Try and detect Let's Encrypt root related errors and show helpful hints
- If sending a event fails, don't try to send a transaction
- Only show generic "DSN is broken" message if the SDK emitted no errors
- Return a unix exit code depending on if the command succeeded or failed

And a bunch more I am probably forgetting. The inner workings have changed but the result has not and also the way they are interacted with have not changed so this should be a very transparent change and just be a glow up for some edge cases.

This should hopefully cut down on boilerplate support messages when setting up the Laravel SDK.

Fixes #521


sforward:
The added commands will not work with Laravel 6 because of the usage of `self::SUCCESS` and `self::FAILURE`.

stayallive:
@sforward good catch, I failed to check that, I'll replace them 👍

PR #524, opened 2021-11-16
Publish command: prompt for test after publishing config and .env

### LINKED ISSUE #523: Publish command doesn't publish config/env when successfully sending a test event/transaction
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
Laravel SDK 2.10.1

### Steps to Reproduce

1. Run `php artisan sentry:publish --dsn={{DSN}}`
2. Answer `yes` to `Enable Performance Monitoring?`
3. Answer `yes` to `Want to send a test event & transaction?`

### Expected Result

The test transaction is sent AND the config is published along with the .env entries.

### Actual Result

Only the test transaction is sent. The config and .env entries only get published when answering `no` to `Want to send a test event & transaction?`

BenSampo:
Resolves #523

### DIFF
--- a/src/Sentry/Laravel/Console/PublishCommand.php
+++ b/src/Sentry/Laravel/Console/PublishCommand.php
@@ -72,19 +72,19 @@ class PublishCommand extends Command
             $env['SENTRY_TRACES_SAMPLE_RATE'] = '0';
         }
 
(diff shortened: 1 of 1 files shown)


stayallive:
I'll explain further in #523.

Issue #449, opened 2021-02-04
Normalized route names when using Lumen


giggsey:
Hi,

When using Tracing in Lumen, it's unable to get the Route Name.

$request->route() returns an array instead of a Route object. This could be used instead to get the same information as full Laravel.

stayallive:
I'm happy to accept a PR fixing this if you are up for it, the initial performance release was and is focussed on Laravel at the moment 👍

mnbp1371:
Hello, I need to run sentry captureException in queue . do you Have any tutorial?

stayallive:
@mnbp1371 no we do not, but running it in the queue is the same as not in the queue, works anywhere in your project. 

But please open up a new issue with what you want to do and what is not working so we can try to help you out. 

But this is not the issue, or at least I don't think it's related to your question.

migralito:
@stayallive cloud you please give me an overview of the steps to follow for merging the PR? so I can make myself with an expectation whether this is ready to merge or not, and the time it will take for this to make to the following release.
Thank you

stayallive:
@migralito I needed some time to check this out.

Unfortunately I am able to break it with a route like: `{type}/user` when called with the URL: `user/user`, it results in `{type}/{type}`.

I know this is maybe a stretch but I'm sure there are some better examples to think of where this could happen.

I was also missing the route name and action that I've added back. But I'm not 100% sure about the above issue yet.

PR #535, opened 2022-01-18
Normalized route names when using Lumen

migralito:
Hi, I found in our projects that when using Lumen, transaction names are sent "as is" in the URL. This makes the problem of many many different transactions because, for example, a /songs/{id}/play endpoint would appear as /songs/1/play, /songs/2/play, etc etc etc.
With this, we can make the transaction names to be reported as /songs/{id}/play, fixing this problem for Lumen users.


migralito:
I've just seen that we can link this PR to [this issue](https://github.com/getsentry/sentry-laravel/issues/449)

stayallive:
@migralito I needed some time to check this out.

Unfortunately I am able to break it with a route like: `{type}/user` when called with the URL: `user/user`, it results in `{type}/{type}`.

I know this is maybe a stretch but I'm sure there are some better examples to think of where this could happen.

I was also missing the route name and action that I've added back. But I'm not 100% sure about the above issue yet.

migralito:
> @migralito I needed some time to check this out.
> 
> Unfortunately I am able to break it with a route like: `{type}/user` when called with the URL: `user/user`, it results in `{type}/{type}`.
> 
> I know this is maybe a stretch but I'm sure there are some better examples to think of where this could happen.
> 
> I was also missing the route name and action that I've added back. But I'm not 100% sure about the above issue yet.

Hi @stayallive ,
I understand that this has some drawbacks. What happens is that this is the way I found to reconstruct the URL. It might not be bulletproof, but is much much better to _maybe_ have some URLs incorrectly named than having IDs and other parameters in the operation name. Currently, the APM is just not giving us the chance of having a representation of the operation because it breaks it down to hundred, maybe thousand of transactions.
In the event of incorrectly infering the name of the transaction, or merging two different transactions into one, there's always the chance of fixing it by customizing the transaction name. Without this, customizing the transaction name is the only way to have a non-broken APM transaction set

migralito:
Hi @stayallive 
Any other news regarding this?
Thanks

stayallive:
Hi @migralito, apologies for this taking so long :)

I have refactored it a bit more and added tests, how does it look to you now?

Issue #536, opened 2022-01-24
Use $container->call for `traces_sampler`


pionl:
👋 all,

It would be great if the `traces_sampler` would not require `static` method but use $container->call to call provided `sample` method in provided class. Why?

- Non static method in class -> we can use `SentryService` convention
- We can use container dependency injection (now we need to use `app`).
- It would be easier to test the "sample" method via PHPUnit test (without Laravel app).

From this code:

```php
class SentryService
{
    /**
     * Sample method required by Sentry SDK.
     *
     * @param SamplingContext $context
     *
     * @return float
     * @throws BindingResolutionException
     */
    public static function sample(SamplingContext $context): float
    {
        $app = app();
        /** @var Guard $guard */
        $guard = $app->make(Guard::class);;
        /** @var Request $request */
        $request = $app->make(Request::class);

        if ($context->getParentSampled()) {
            // If the parent transaction (for example a JavaScript front-end)
            // is sampled, also sample the current transaction
            return 1.0;
        }

        if ($guard->check() === false) {
            return 0.0;
        }

        if ($request->cookie('performance', false) !== false) {
            return 1.0;
        }

        return 0.1;
    }
}
```

to this code

```php
class SentryService
{
    /**
     * Sample method required by Sentry SDK.
     *
     * @param SamplingContext $context
     * @param Guard           $guard
     * @param Request         $request
     *
     * @return float
     */
    public static function sample(SamplingContext $context, Guard $guard, Request $request): float
    {
        if ($context->getParentSampled()) {
            // If the parent transaction (for example a JavaScript front-end)
            // is sampled, also sample the current transaction
            return 1.0;
        }

        if ($guard->check() === false) {
            return 0.0;
        }

        if ($request->cookie('performance', false) !== false) {
            return 1.0;
        }

        return 0.1;
    }
}
```

Thank you for considering. If needed, I can make PR.

Have a great day,

stayallive:
I might be missing something here but you still have a `static` method in your `SentryService` now right? 😄  
But I think I get what you intended 👍 

There is currently no special handling for the `traces_sampler` config option, it's handled by the underlying SDK so I'm not sure we can (without breaking existing code) override that and let the Laravel container handle it, maybe it is as simple as wrapping the option in a callback ourselfs using the container but give me some time to think about this and the effect it might have, otherwise it might have to wait for a next major version which I'd rather not do.

pionl:
Hi, thanks for answering. It should not be a problem.

[...]

PR #537, opened 2022-01-28
traces_sampler - support dependency injection

pionl:
Use container dependency injection for provided callable / class method traces_sampler

- Supports static / non-static method in the class
- Supports callables

Related to to #536

Is it possible write tests targeting Lumen? Not use if it works.

### DIFF
--- a/src/Sentry/Laravel/ServiceProvider.php
+++ b/src/Sentry/Laravel/ServiceProvider.php
@@ -141,7 +141,7 @@ class ServiceProvider extends BaseServiceProvider
             Integration::setControllersBaseNamespace($userConfig['controllers_base_namespace']);
         }
 
-        $this->app->bind(ClientBuilderInterface::class, function (Application $app) {
+        $this->app->bind(ClientBuilderInterface::class, function ($app) {
             $basePath   = base_path();
             $userConfig = $this->getUserConfig();
 
@@ -159,6 +159,7 @@ class ServiceProvider extends BaseServiceProvider
 
             if (isset($options['traces_sampler']) === true) {
                 /** @var CallableProxy $proxy */
+                /** @var Application $app */
                 $proxy = $app->make(CallableProxy::class);
 
                 $options['traces_sampler'] = $proxy->proxyTraceSampler($options['traces_sampler']);
--- a/src/Sentry/Laravel/ServiceProvider.php
+++ b/src/Sentry/Laravel/ServiceProvider.php
@@ -2,6 +2,7 @@
 
 namespace Sentry\Laravel;
 
+use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Contracts\Http\Kernel as HttpKernelInterface;
 use Illuminate\Foundation\Application as Laravel;
 use Illuminate\Foundation\Http\Kernel as HttpKernel;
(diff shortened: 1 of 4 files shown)


src/Sentry/Laravel/ServiceProvider.php
```diff
@@ -139,7 +141,7 @@ protected function configureAndRegisterClient(): void
             Integration::setControllersBaseNamespace($userConfig['controllers_base_namespace']);
         }
 
-        $this->app->bind(ClientBuilderInterface::class, function () {
+        $this->app->bind(ClientBuilderInterface::class, function (Application $app) {
```
pionl, line 144: What lowest laravel we are targeting? Should be always Application contract

pionl:
All green :)

Snímek obrazovky 2022-01-28 v 12 13 31

Snímek obrazovky 2022-01-28 v 12 14 28

src/Sentry/Laravel/Tracing/CallableProxy.php
```diff
@@ -0,0 +1,62 @@
...
+
+    public function __construct(Container $container)
+    {
+        $this->container = $container;
+    }
+
+    /**
+     * @param array|callable|null $tracesSampler
+     *
+     * @return array|callable
```
stayallive, line 27:
```suggestion
     * @param array|callable|null $tracesSampler
     *
     * @return array|callable|null
```

src/Sentry/Laravel/Tracing/CallableProxy.php
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public function proxyTraceSampler($tracesSampler)
```
stayallive, line 29: Is there a way to make this more generic? So we can proxy any callable? I would like to be able to apply this to all possible callback enabled options and this is highly specific for the `traces_sampler` at the moment.

stayallive (changes requested):
Few comments and questions.

Also check out the `composer.json` we still support PHP 7.2 and Laravel 5.

https://github.com/getsentry/sentry-laravel/blob/5ae05c8b41bff5ffac7bff44cdb063d0f9cda292/composer.json#L24-L25

I think you used some PHP 8 specific features which is a no-go at the moment.

And no testing against Lumen has proofed very difficult so unless doing it manually not possible in the test suite at the moment.

pionl:
1. Did not use PHP 8 features i think - the problem is that Laravel 8 and above made possible to use `[class, method]`
2. Maybe disable the feature on older versions?

src/Sentry/Laravel/Tracing/CallableProxy.php
pionl: Im gona remove it if there are not tests for this part :)

src/Sentry/Laravel/Tracing/CallableProxy.php
pionl: That was my thought too. but the problem is the received parameter. Im gona try to prevent the container changing the "context"

pionl: So, what I thought. We need to map the arguments with either name => $value, or class => $value. But I have not found a method that will return arguments of the function with `$variable` name. Depedency injection does not support `ordered` parameters.


```
if (is_callable($tracesSampler) === true) {
            return function () use ($tracesSampler) {
                return $this->container->call($tracesSampler, func_get_args());
            };
        }
```


We can do not bullet proof solution, but working solution:

-> loop all arguments and if it is a object, get the class and store it to a map (class => $value)
-> take this map and pass it to the `call` method.

stayallive:
@pionl sorry Martin, I don't think I am going to feel comfortable adding this code in just to remove a little extra work / lines on the side of the user for these callables by using the `app()` method or other ways of container injection.

I've tried to make it more generic and less convoluted but I don't think it's happening as of this moment, especially considering the amount of Laravel/Lumen en PHP versions supported.

We might revisit this in the future but for now I am going to close this. 

Thank your for the effort you put into it and suggesting it originally!

PR #543, opened 2022-02-14
Fix deprecation notice in route name extraction

X-Coder264:
We just updated one of our projects from Laravel 8 and PHP 8.0 to Laravel 9 and PHP 8.1 and immediately

`str_ends_with(): Passing null to parameter #1 ($haystack) of type string is deprecated in /code/vendor/laravel/framework/src/Illuminate/Support/Str.php on line 247`

started popping up on Sentry (as we have the deprecation logger configured for the app).

Turns out there is a code path here that triggers that deprecation (specifically the code path when the route has no name so Laravel auto generates the name for the route).

This PR fixes that deprecation and adds tests for the method that this PR touched.


stayallive:
@X-Coder264, nice catch! We indeed missed that.

I did some refactors and added some more tests, I hope you do not mind 😄 

(CI seems to have some temporary issues at the moment, will retry them later)

X-Coder264:
@stayallive Since you went and refactored a bunch of code, just one more thing that I've noticed.

`public static function extractNameForRoute(Route $route): ?string`

The return type is nullable even though `null` can never be returned from this method:

```php
        // /someaction // Fallback to the url
        if (empty($routeName) || $routeName === 'Closure') {
            $routeName = '/' . ltrim($route->uri(), '/');
        }

        return $routeName;
```

If $routeName is null `empty($routeName)` will return `true` and the if will be executed turning `$routeName` into a string so the return type should be `string`, not `?string`.

X-Coder264:
You changed the return type of `extractNameForActionRoute` though, instead of `extractNameForRoute`. I'm not sure if `extractNameForActionRoute` can or cannot return `null`.

stayallive (approved):

stayallive:
Thanks for the PR and the feedback @X-Coder264, 2.11.1 with this fix should be released soon 💪

X-Coder264:
Cool, thank you for such a quick review / merging / release tagging process.

PR #557, opened 2022-05-27
Allow extending the Laravel specific config options

aand18:
Adding custom config keys to `/config/sentry.php` causes the Sentry SDK to throw an error.

This change allows hiding the new config from the Sentry SDK by adding custom config values to the `LARAVEL_SPECIFIC_OPTIONS` array in a custom service provider:

```php
//    /config/sentry.php

return [
    'custom_config_key' => 'some value',      // adding this caused the SDK to throw an error

    // the rest of Sentry's config
    // ...
];
```

```php
class App\Providers\CustomSentryServiceProvider extends Sentry\Laravel\ServiceProvider
{
    protected const LARAVEL_SPECIFIC_OPTIONS = [
        'custom_config_key',

        // other Laravel-specific config
        // ...
    ];
}
```

### DIFF
--- a/src/Sentry/Laravel/ServiceProvider.php
+++ b/src/Sentry/Laravel/ServiceProvider.php
@@ -26,7 +26,7 @@ class ServiceProvider extends BaseServiceProvider
     /**
      * List of configuration options that are Laravel specific and should not be sent to the base PHP SDK.
      */
-    private const LARAVEL_SPECIFIC_OPTIONS = [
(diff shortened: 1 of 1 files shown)


stayallive:
Oh, I'm really sorry this went under the radar! I'm taking a look at it now 🤘

stayallive (approved):
Simple and effective, thanks! 👍

aand18:
> Simple and effective, thanks! 👍

I guess a getter and setter would have been the Right Thing™, but this is indeed cleaner :)

PR #570, opened 2022-09-23
fix: Update span ops

cleptric:
To unblock the work on Performance Issues, we have to update some span ops.
https://github.com/getsentry/sentry/discussions/37083

### DIFF
--- a/src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
+++ b/src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
@@ -63,7 +63,7 @@ class LighthouseIntegration implements IntegrationInterface
         }
 
         $context = new SpanContext;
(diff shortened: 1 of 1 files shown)


AbhiPrasad (approved):

stayallive (changes requested):
Maybe I am misunderstanding what the `db` prefix does or mean but this change makes little sense to me. A GraphQL span is not a database query, it can contain database queries of course but itself isn't one. A GraphQL query/request might not even touch the database.

I have an example of a current trace generated by this SDK which shows a typical GraphQL request:

image

Maybe the `db.sql.query` needs to be called differently to be useful to Sentry instead?

cleptric:
@k-fish Any thoughts on this?

cleptric:
After some discussions with @AbhiPrasad and @jan-auer, we'll get back to the drawing board.

PR #573, opened 2022-09-29
fix: Status code not populated on transaction if response did not inherit from the Laravel Response class

ydwis40:
Fixing terminate method if return is JsonResponse instance


stayallive (approved):

cleptric (approved):
Thanks guys!

PR #575, opened 2022-09-30
fix: Broken `SetRequestMiddleware` on Laravel < 6.0

### LINKED ISSUE #473: Class 'Symfony\\Bridge\\PsrHttpMessage\\Factory\\DiactorosFactory' not found
Hi,

After updating sentry-laravel from 2.3.1 to 2.4 exception capture stops working.
That's connected to using symfony/psr-http-message-bridge v2 (v1 does not throw error).

stayallive:
Release 2.13.0 broke the `SetRequestMiddleware` on Laravel < 6.0 when `symfony/psr-http-message-bridge:^1.0` and `zendframework/zend-diactoros` was missing. One of those we had listed in our suggest but still we should never break of course.

So I've opted to backport the current code Laravel uses to construct a PSR-7 request, we already require the needed dependencies for that so I'm not sure why we weren't doing this before... but better late than never :)

Also snuk in removing a constructor argument that was not used 👀 

Fixes #473.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fix not listening to queue events because `QueueManager` is registered as `queue` in the container and not by it's class name (#568)
 - Fix status code not populated on transaction if response did not inherit from `Illuminate\Http\Response` like `Illuminate\Http\JsonResponse` (#573)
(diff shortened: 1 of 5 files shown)


cleptric (approved):
Good stuff! 😀

PR #576, opened 2022-09-30
Upgrade php-cs-fixer dependency

stayallive:
Just internal, no real changes.

### DIFF
--- a/.github/workflows/cs.yaml
+++ b/.github/workflows/cs.yaml
@@ -18,7 +18,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '7.4'
+          php-version: '8.1'
 
       - name: Install dependencies
         run: composer update --no-progress --no-interaction --prefer-dist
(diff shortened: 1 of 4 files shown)


.github/workflows/cs.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '7.4'
+          php-version: '8.1'
```
cleptric, line 21: Might this point out to use PHP language features not available in `7.4`?
Our is our ruleset properly configured for this?

stayallive: No, CS is just that, CS AFAIK it doesn't do any language feature pointing out unfortunately.

cleptric (approved):

.github/workflows/cs.yaml
stayallive: Let's expore that in a different PR, this was mostly done so you can use Laravel 9 (latest) in local development :)

PR #579, opened 2022-10-06
Drop Laravel Lumen support

stayallive:
The Laravel Lumen framework is no longer supported or even endorsed

> Note: In the years since releasing Lumen, PHP has made a variety of wonderful performance improvements. For this reason, along with the availability of [Laravel Octane](https://laravel.com/docs/octane), we no longer recommend that you begin new projects with Lumen. Instead, we recommend always beginning new projects with [Laravel](https://laravel.com/).

^ this is from the official Laravel Lumen docs site: https://lumen.laravel.com, it's install usage is also declining and never really took of: https://packagist.org/packages/laravel/lumen/stats.

Since we have to maintian compatibility we are held back in some of the things we want to do with tracing and the framework is also currently not tested or possible to test together with Laravel.

Going forward in 3.x of the `sentry/sentry-laravel` package we want drop support for Lumen.

Of course Lumen users can keep using version `2.14.x` and below like they are now.


cleptric (approved):
To add the numbers, we're looking at less than `400` daily installs for `laravel/lumen`, compared to over `25.000` for `laravel/laravel`.

LGTM! :shipit:

Issue #562, opened 2022-08-01
Raw URLs sent as Transaction


smeubank:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) 

Which SDK and version?
sentry-laravel 2.12.0

### Steps to Reproduce

Sending transactions

### Expected Result

High quality transactions

### Actual Result

for sentry laravel (2.12.0) we're getting raw URLs and bad transactions. we noticed:
- OPTIONS has raw URLs. probably shouldn't be recorded to begin with
- there's a bunch of transactions reported without http.method
- 404s are missing the http.status, and those also come with raw URLs

smeubank:
for more context, to improve the performance product experience, we are working to improve transaction naming

meaning that in product we don't have a ton of unique transactions because of poor parameterization from different platforms

Meaning the org slug `…/sentry/…` becomes `…/:orgId/… `

1. all web SDKs should now send source: [dev spec](https://develop.sentry.dev/sdk/event-payloads/transaction/#transaction-annotations)
2. [DSC spec](https://develop.sentry.dev/sdk/performance/dynamic-sampling-context/#note-on-low-quality-transaction-names) regarding transaction name quality

smeubank:
and sentry-symphony sends URLs for everything

smeubank:
@jan-auer is there some data we can share with @stayallive regarding 

> there's a bunch of transactions reported without http.method

for option we we agreed to exclude similar to other SDKs it is not useful

cleptric:
@stayallive We should take this into consideration for the DS rollout.

cleptric:
Done in https://github.com/getsentry/sentry-laravel/pull/583 and to be released in v3.0.0

PR #580, opened 2022-10-06
Improve tracing

stayallive:
We used to set the transaction name as late as possible to make sure the routing information is available to us. However this doesn't work well with generating the new `baggage` so we now do it as early as Laravel allows us instead.

Few things I like to improve here still:

- [x] Use the `RouteMatched` event for setting the transaction name
- [x] Do not trace hard-404's (no route)
- [ ] Do not trace `HEAD` requests

Discussion points:

- Do we want to remove the route name (`some.route`) and controller class (`SomeController@someAction`) as options for transaction names and only keep the "route" version of the URL (`some/{route}`)?
- Do we allow people to opt-in `HEAD` request and hard-404 tracing?


src/Sentry/Laravel/Tracing/Middleware.php
```diff
@@ -108,12 +102,17 @@ private function startTransaction(Request $request, HubInterface $sentry): void
             $request->header('baggage', '')
         );
 
+        $requestPath = '/' . ltrim($request->path(), '/');
+
         $context->setOp('http.server');
+        $context->setName($requestPath);
+        $context->setSource(TransactionSource::url());
```
cleptric, line 109: `m:` This should be `TransactionSource::route()`

cleptric (commented):
The main reason for this change is to properly support the use case where the Laravel SDK is head of trace, setting the `baggage` in the meta tag. Having the transaction name figured out before the view is rendered is something we have to do.

Setting unified transaction names based on the route version sounds like a good idea.

Worth checking what happens in https://github.com/getsentry/sentry-python/pull/1661#issuecomment-1268448333. This will go out as a new major version anyway, but maybe we get some insights onto how many dashboards could be affected.

As mentioned in https://github.com/getsentry/sentry-laravel/issues/562, I would wait for changes to how we instrument `HEAD/OPTIONS` requests for now.

A way to enable instrumentation on 404s sounds good to me.

cleptric (approved):

PR #581, opened 2022-10-06
Drop Laravel 5.x support

stayallive:
Going forward in 3.x we intend to drop support for Laravel versions older than 6.x (LTS).

This means dropping the currently supported versions `5.0-5.8` and we only support Laravel `6+` where `6.0` was released on September 3, 2019.

Usage is low enough that we think this can be done at this time: https://packagist.org/packages/laravel/laravel/stats#major/all.

Of course users of those Laravel versions can keep using `2.14.x` and below like they are now.

For a bit of background when those versions lost support: https://laravelversions.com/en.

This is a draft for discussion and because of the big cleanup it needs another once over to make sure we removed all version checks and "hacks" to support these older versions.


cleptric (approved):
> +143 −388
Nice!

cleptric:
@stayallive We should also update the README https://github.com/getsentry/sentry-laravel#laravel-version-compatibility

PR #583, opened 2022-10-07
Remove extracting route name or controller for transaction names

stayallive:
After some discussion it might be best if we only show the `uri` (parameterized URL) as a transaction name instead of defaulting to the developer provided route name or controller class FQN.

This is slated for 3.x and technically not a breaking change but it will change almost all transaction names for most users.


denisdulici:
After this change, different and unrelated actions/events will be grouped under the same transaction name. For example, the `index` (GET) action and `store` (POST) action:

https://laravel.com/docs/9.x/controllers#actions-handled-by-resource-controller

I guess this will cause a lot of confusion.

stayallive:
They are not, they are still grouped by their HTTP method.

image

denisdulici:
Hmm, you're right. While they're not as cool/clear as before, they're not mixed:

Screenshot_1095

Thanks for the prompt response 🙏

PR #585, opened 2022-10-10
Add tracing span for Laravel HTTP client requests

### LINKED ISSUE #547: Span trace for Laravel HTTP Client Requests
Hey!

Would be awesome to add support for tracing http requests made via the laravel HTTP client. Should be fairly easy as events are fired.

https://laravel.com/docs/9.x/http-client#events

stayallive:
There were no obvious global hooks to potentially inject our `baggage` header into the request, but we could ship a simple middleware that accomplishes that in the future.

Fixes #547.


cleptric (approved):
Awesome stuff :shipit:

PR #586, opened 2022-10-10
Simplify Sentry meta tag retrieval

stayallive:
This simplifies echo'ing the Sentry meta for the end-user by just needing to call:

```php
Sentry\Laravel\Integration::sentryMeta();
```

This makes it easier than needing (and documenting) to do:

```php
Sentry\Laravel\Integration::sentryTracingMeta();
Sentry\Laravel\Integration::sentryBaggageMeta();
```

We could even drop the "seperate" methods or make them private / internal. But keeping them around for now makes this a non-bc change which saves one more step users _need_ to take when upgrading (but might be worth still... to be discussed).

### DIFF
--- a/src/Sentry/Laravel/Integration.php
+++ b/src/Sentry/Laravel/Integration.php
@@ -216,6 +216,17 @@ class Integration implements IntegrationInterface
 
     /**
      * Retrieve the meta tags with tracing information to link this request to front-end requests.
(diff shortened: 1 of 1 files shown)


cleptric (approved):
Good stuff :shipit:

We could leave the methods and point people to `sentryMeta` via docs.
This makes the upgrade less painful imo.

Issue #595, opened 2022-10-14
The SDK assumes a `trace_sample_rate` of 0 means we don't want to trace


AAllport:
### Environment

When running v2.13, setting the `SENTRY_TRACES_SAMPLE_RATE` to `0.0` will disable the tracing integration.
Whilst this is understandable when serving the frontend via Laravel, this does not work well when using Laravel as an API only.

Our use case has a VueJs frontend, and a Mobile client.
We are enabling performance monitoring and initially hoped to have the sampling decision set on the VueJs frontend for this initial phase.

Due to how couldHavePerformanceTracingEnabled is implemented, a 0 value causes tracing not to be loaded.
https://github.com/getsentry/sentry-laravel/blob/acbf23632582ec058a8fb51be2c5bc3a2283a52c/src/Sentry/Laravel/BaseServiceProvider.php#L47-L52

### Steps to Reproduce

1. Enable the SDK with a DSN and 0 Sample rate
2. Have a front end with a positive sampling decision to make a request to the Laravel application.

### Expected Result

A backend trace should be sent to Sentry with its parent set to the frontend id

### Actual Result

The backend tracing integration was not enabled, and no trace was sent

stayallive:
For reference this issue was sparked by a conversation in the Sentry Discord: https://discord.com/channels/621778831602221064/621964286952210432/1030461140913897482.

This is assumed to be undefined or at the very least unclear behaviour of the SDK and we should decide how to proceed because it would still be nice to have the SDK not do all the tracing bits if the user is never going to use it, but our very naive heuristic at the moment is obviously also not ideal or correct behaviour.

A workaround for now could be defining your own `trace_sampler` callback:

```php
'traces_sampler' => function (\Sentry\Tracing\SamplingContext $context): float { return 0.0; },
```

However this is not ideal of course but does workaround the issue until we figure out how to fix it correctly.

cleptric:
Just confirmed, we have to respect the sampling decision of the HEAD SDK, so we should instrument this regardless.

cleptric:
We should do this right away in a bug-fix release.

cleptric:
@AAllport First of all, thanks for reporting this. We actually discovered the underlying issue to be in our generic PHP SDK and will need to fix this there first.

AAllport:
Oh wow, didn't realise there was an issue with both of the SDKs.
[...]

PR #587, opened 2022-10-11
[WIP] Laravel SDK version 3.0.0

cleptric:
This will end up being a new major version of our Laravel SDK.

**Remaining ToDos**

- [x] Add support for Dynamic Sampling https://github.com/getsentry/sentry-symfony/pull/661
- [ ] Set the correct transaction name early in the request cyclehttps://github.com/getsentry/sentry-laravel/pull/580
  - [ ] Unify transaction names https://github.com/getsentry/sentry-laravel/pull/583
  - [ ] Simplify setting the correct `<meta>` tags for tracing and Dynamic Samling https://github.com/getsentry/sentry-laravel/pull/586
  - [ ] Automatically instrument the Laravel HTTP client https://github.com/getsentry/sentry-laravel/pull/585
  
**Deprecations**

- [ ] Drop Laravel Lumen support https://github.com/getsentry/sentry-laravel/pull/581
- [ ] Drop Laravel version 5.x support https://github.com/getsentry/sentry-laravel/pull/581

**Docs and product**

- [ ] Enable Dynamic Sampling for the Laravel SDK https://github.com/getsentry/sentry/pull/39670/files
- [ ] Adjust docs for Laravel Dynamic Sampling https://github.com/getsentry/sentry-docs/pull/5601

PR #600, opened 2022-10-18
Always start performance tracing

stayallive:
As mentioned in #595, we are incorrectly "guarding" against enabling performance tracing when no sample rate or sampler callback is set. This removes that check and always enables performance tracing.

This has minimal impact on the users application since we always check if there is an active transaction/span before doing any real work so most code becomes a no-op if the transaction is not sampled.

This should be merged to `master` after #587.

This will also have no real work effect until getsentry/sentry-php#1405 has been solved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Remove incorrect checks if performance tracing should be enabled and rely on the transaction sampling decision instead (#600)
+- 
 ## 3.0.0
 
 **New features**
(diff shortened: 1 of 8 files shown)


cleptric (changes requested):
`h:` We have to bump `sentry/sentry` to `^3.10`.
And while we're at it, `sentry/sdk` to `^3.3`.

`h:` `Middleware::hydrateResponseData()` throws an `Call to a member function setHttpStatus() on null` error, when `traces_sample_rate` is set to`0`.

cleptric (approved):

PR #601, opened 2022-10-18
ref: Remove examples

cleptric:
I don't think these examples are valuable anymore.


stayallive (approved):
`l`: You can also cleanup the `.gitattributes`.

Good change, 🚢 it!

Issue #167, opened 2018-12-09
Git 'release' not working when running artisan commands outside of root directory


ryzr:
Please see https://github.com/getsentry/sentry-laravel/pull/51

This pull request fixes the issue, but the merge seems to have snuck out
https://github.com/getsentry/sentry-laravel/blame/master/src/Sentry/SentryLaravel/config.php#L7

PR #603, opened 2022-10-22
Support using SENTRY_RELEASE environment variable

challgren:
According to the [PHP SDK docs](https://docs.sentry.io/platforms/php/configuration/options/#release) a user can set the `SENTRY_RELEASE` environment variable. But with the Laravel specific package the `release` option is commented out and not using the default value.

### DIFF
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -4,8 +4,9 @@ return [
 
     'dsn' => env('SENTRY_LARAVEL_DSN', env('SENTRY_DSN')),
 
-    // capture release as git sha
-    'release' => env('SENTRY_RELEASE', trim(exec('git --git-dir ' . base_path('.git') . ' log --pretty="%h" -n1 HEAD'))),
+    // The release version of your application
+    // Example with dynamic git hash: trim(exec('git --git-dir ' . base_path('.git') . ' log --pretty="%h" -n1 HEAD'))
+    'release' => env('SENTRY_RELEASE'),
 
     // When left empty or `null` the Laravel environment will be used
     'environment' => env('SENTRY_ENVIRONMENT'),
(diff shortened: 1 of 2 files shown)


challgren:
Related https://github.com/getsentry/sentry-laravel/pull/51 and https://github.com/getsentry/sentry-laravel/issues/167

stayallive:
What exactly is not working with setting `SENTRY_RELEASE` in your `.env`? It works just fine without any changes.

Also this PR enables running a shell command by default which we are not doing for a few reasons (mainly security and performance).

config/sentry.php
```diff
@@ -5,7 +5,7 @@
     'dsn' => env('SENTRY_LARAVEL_DSN', env('SENTRY_DSN')),
 
     // capture release as git sha
-    // 'release' => trim(exec('git --git-dir ' . base_path('.git') . ' log --pretty="%h" -n1 HEAD')),
+    'release' => env('SENTRY_RELEASE', trim(exec('git --git-dir ' . base_path('.git') . ' log --pretty="%h" -n1 HEAD'))),
```
stayallive, line 8: I would be fine with this, but I am not with executing a shell command on our users machines by default, even though it looks harmless enough. Not everyone is using git or config caching which can cause issues.

```suggestion
    'release' => env('SENTRY_RELEASE'),
```

stayallive (changes requested):
@cleptric I'm not sure we even want to keep the `git` command example in the config file, it's been around since forever and it's been kept but this might be better documented elsewhere because this is not the most efficient way to achieve this anyway.

config/sentry.php
cleptric: I would also rather have this removed and go ahead with your suggestion here.

cleptric (approved):

PR #608, opened 2022-11-09
Introduce `captureUnhandledException`

### LINKED ISSUE #607: Correctly mark unhandled exceptions as `handled: false`
We exposed the `ExceptionMechanism` via the `EventHint`, so we can now correctly set the handled property in the SDK's error handler.

Replace all calls to `captureException()` with

```
$hint = EventHint::fromArray([
    'exception' => $exception,
    'mechanism' => new ExceptionMechanism(ExceptionMechanism::TYPE_GENERIC, false),
]);

captureEvent(Event::createEvent(), $hint);
```

https://github.com/getsentry/sentry-php/pull/1416

stayallive:
Currently we ask users to implement this in their error handler:

```php
public function register()
{
    $this->reportable(function (Throwable $e) {
        if (app()->bound('sentry')) {
            app('sentry')->captureException($e);
        }
    });
}
```

However going forward we are going to change this to:

```php
public function register()
{
    $this->reportable(function (Throwable $e) {
        \Sentry\Laravel\Integration::captureUnhandledException($e);
    });
}
```

The old way will still work but the "new way" will make sure exception are correctly marked as unhandled in the Sentry UI.

Fixes #607 (TODO: Docs PR)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Add new method to report unhandled exceptions from the Laravel error handler (#608)
+
 ## 3.0.1
 
 - Remove incorrect checks if performance tracing should be enabled and rely on the transaction sampling decision instead (#600)
--- a/src/Sentry/Laravel/Integration.php
+++ b/src/Sentry/Laravel/Integration.php
@@ -3,9 +3,14 @@
 namespace Sentry\Laravel;
 
 use Illuminate\Routing\Route;
+use Sentry\EventHint;
+use Sentry\EventId;
+use Sentry\ExceptionMechanism;
 use Sentry\SentrySdk;
 use Sentry\Tracing\TransactionSource;
+use Throwable;
 use function Sentry\addBreadcrumb;
+use function Sentry\captureEvent;
 use function Sentry\configureScope;
 use Sentry\Breadcrumb;
 use Sentry\Event;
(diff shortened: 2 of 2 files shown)


stayallive:
I did thought about that, however, it's not certain that we have a unhandled exception there since a user could also just log an exception like they would use captureException, so it wouldn't be a 100% correct. WDYT?

cleptric:
Fair point, so let's leave it as is.

cleptric (approved):

devfrey:
After switching from `captureException()` to `captureUnhandledException()` I'm no longer seeing any additional context in Sentry. No headers, no user, no breadcrumbs.

I'm using v3.1.0 of `sentry/sentry-laravel`.

stayallive:
@devfrey you are right... fixing that now @ #611!

stayallive:
3.1.1 was just released fixing this, sorry for that!

cleptric:
Our goal was to introduce a distinction between errors sent via `captureException()` `handled: true` and errors caught by the global error handler of the framework `handled: false`.

https://laravel.com/docs/9.x/errors#the-report-helper

```
public function isValid($value)
{
    try {
        // Validate the value...
    } catch (Throwable $e) {
        \Sentry\captureException($e); // handled: true
 
        return false;
    }
}
```

It's a bit odd that `report()` behaves that way. So we may have to make a decision on how to handle this going forward.
Having `handled: false` is highly valuable, as it highlights the errors with higher priority IMO.

cleptric:
@devfrey @axlon [3.1.2](https://github.com/getsentry/sentry-laravel/releases/tag/3.1.2) was just released that properly sets the `handled` property when using `report()`.

cleptric:
Yes, `report_if()` and `report_unless()` are also set to `handled: true` 🙂

cleptric:
Yes, [`report_if()`](https://laravel.com/docs/9.x/helpers#method-report-if) and [`report_unless()`](https://laravel.com/docs/9.x/helpers#method-report-unless) are also set to `handled: true` 🙂

PR #616, opened 2022-11-21
Set `traces_sample_rate` to `null` by default

stayallive:
- [x] Pin `sentry/sentry` to version including: getsentry/sentry-php#1428

(missing `sentry/sentry` release is failing the tests)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Set `traces_sample_rate` to `null` by default (#616)
+
 ## 3.1.1
 
 - Fix missing scope information on unhandled exceptions (#611)
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     "require": {
         "php": "^7.2 | ^8.0",
         "illuminate/support": "^6.0 | ^7.0 | ^8.0 | ^9.0",
-        "sentry/sentry": "^3.10",
+        "sentry/sentry": "^3.12",
         "sentry/sdk": "^3.3",
         "symfony/psr-http-message-bridge": "^1.0 | ^2.0",
         "nyholm/psr7": "^1.0"
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -51,9 +51,10 @@ return [
         'missing_routes' => false,
     ],
 
-    // @see: https://docs.sentry.io/platforms/php/configuration/options/#send-default-pii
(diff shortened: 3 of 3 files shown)


cleptric (approved):

negoziator:
@stayallive I really like this change, since the default behavior is now not to send trace samples ( which can get pretty expensive quickly.... )

However .. i'm just curious why it wouldn't be enough to just do it like this instead of the explicit type check for null` ?

```php
'traces_sample_rate' => env('SENTRY_TRACES_SAMPLE_RATE') ?: (float)env('SENTRY_TRACES_SAMPLE_RATE'),
```

I'm maybe overseeing something, but by doing it like this - it seems to me like there's a reason for being very explicit ( and also more complex to read )

stayallive:
To be clear, there were never any traces sent by default, except when you had a front-end library instructing to send traces which caught some users off guard (understandably). This is why this was changed, but we didn't send traces by default without any input (from the front-end or setting this to > 0). Just to be clear to future readers 😃 

Some would argue your example is harder to read, that is code style preference (especially your earlier edit with `?;`). However, as you suspected there is an functional difference too. Your example doesn't check for null but for truthy values, a string or `true` or something else entirely is also truthy and the PHP SDK will fail on that so we check a little more explicit since only `null` and `int` or `float` is acceptable (other types throw a exception which we rather avoid).

stayallive:
So my comment became a bit of a mess because you changed your example (I answered based on the GitHub notification e-mail) sorry about that :)

Your current example doesn't allow setting the `traces_sample_rate` to `0` since that would result in `null` which is different from setting it to `0`. Simply put, you want to be able to set the `traces_sample_rate` anywhere from `0.0` to `1.0` and `null`.

negoziator:
I get your point. Thanks for taking your time to elaborate.

It actually makes sense, that 

if you set it to `0` - then you want the value to be `0.0` 

In my way - when you set it to `0` - that would elaborate to `null` - i get it, that's weird behavior.

Alright. Thank you.

And yes -- sorry for the confusion about the default behavior! 😄

chrillep:
Skärmavbild 2022-11-23 kl  22 19 22
phpstan: Cannot cast mixed to float.
@stayallive is this a false positive?

stayallive:
It's not incorrect but also not something we can fix since the Laravel `env` helper returns `mixed` although in reality it would only return a number or `null` if setup correctly for this key. The old previous value we had should have the same problem.

PR #622, opened 2022-12-08
Change debug trace limit in `makeAnEducatedGuessIfTheExceptionMaybeWasHandled` method

kayw-geek:
The current method to limiting the call stack is not suitable for all users of the Sentry SDK, for example, some users have added additional logic based on Sentry, which can cause the call stack to exceed the limit and result in an unexpected unhandled flag.

### DIFF
--- a/src/Sentry/Laravel/Integration.php
+++ b/src/Sentry/Laravel/Integration.php
@@ -200,7 +200,7 @@ class Integration implements IntegrationInterface
     private static function makeAnEducatedGuessIfTheExceptionMaybeWasHandled(): bool
     {
         // We limit the amount of backtrace frames since it is very unlikely to be any deeper
-        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 10);
+        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 20);
 
         // We are looking for `$handler->report()` to be called from the `report()` function
         foreach ($trace as $frameIndex => $frame) {
@@ -228,7 +228,7 @@ class Integration implements IntegrationInterface
             }
 
             // If we reached this point we can be pretty sure the `report` function was called
-            // and we can can come to the educated conclusion the exception was indeed handled
+            // and we can come to the educated conclusion the exception was indeed handled
             return true;
         }
 


cleptric (commented):
I think the current default is very reasonable.
Can you provide us with some concrete example of where this would not work?

kayw-geek:
> I think the current default is very reasonable. Can you provide us with some concrete example of where this would not work?
OK, I have some custom extra logic in our own private library `laravel-sentry`, so I didn't get to report helper until the 11th call stack.

image
image

kayw-geek:
I don't think this is an individual case. 
and I don't think limiting the number of call stack calls has much significance in terms of performance, because when we find this `report helper` in a loop, it will return and stop the loop.

stayallive:
> I don't think this is an individual case.

You are the first to report this, so unless we have a lot more people chip in, it might be. In a "normal" Laravel application we expect the stack to be around 5-6 long before hitting the report so you are definitly doing a lot of indirection.

Regardless, we limit the amout of stack frames to prevent doing work that is not needed, Sentry threads a fine line between reporting accurate errors and being performant enought to not cause more problems than it's reporting.

That said, I'd be willing to bump it to 20 because I can see 10 might be a little on the thin side, but making it unbounded is not a good idea since not all errors go through report and the loop is not "free" especially with your example where you have 80+ stackframes :)

20 seems to also cover your case and then some, sounds good?

kayw-geek:
> 

Sure, I will change the limit count to 20.

stayallive (approved):
Thanks!

kayw-geek:
Hi @cleptric, Is there any plan to release a new tag recently?

cleptric:
@kayw-geek We'll get something out this week.

kayw-geek:
Thanks!
----- 原始邮件 -----
发件人：Michi Hoffmann ***@***.***>
收件人：getsentry/sentry-laravel ***@***.***>
抄送人："Kay W." ***@***.***>,  Mention ***@***.***>
主题：Re: [getsentry/sentry-laravel] Change debug trace limit in `makeAnEducatedGuessIfTheExceptionMaybeWasHandled` method (PR #622)
日期：2023年01月09日 19点44分


@kayw-geek We'll get something out this week.

&mdash;

kayw-geek:
> @kayw-geek We'll get something out this week.

Thanks!

cleptric:
@kayw-geek release in https://github.com/getsentry/sentry-laravel/releases/tag/3.1.3

kayw-geek:
> 

Thanks, I see it!

PR #624, opened 2022-12-12
ref: Bump CI to PHP 8.2

cleptric:
https://github.com/PHP-CS-Fixer/PHP-CS-Fixer does not work with 8.2 yet, so I did not bump the PHP version there.


stayallive (approved):
💪 (I thought we already did this... good call!)

PR #630, opened 2023-01-10
Add Laravel 10 support

jnoordsij:
This PR aims to add support for Laravel 10, which is [expected to be released at February 7th, 2023](https://laravel.com/docs/master/releases#support-policy).

Open problem: the update to Laravel 10 comes with an update to Monolog v3, which changes the signature on the `AbstractProcessingHandler::write` method. This can be circumvented on PHP ^8.0 by using a union type in the class extension, but this does not work on PHP 7.x as it has no union types.

Possible solutions:
- drop support for older Laravel versions
- drop support for PHP 7.x: see https://github.com/jnoordsij/sentry-laravel/pull/2
- find some sort of polyfill to allow for union types on PHP 7.x

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         php: [ "8.2", "8.1", "8.0", "7.4", "7.3", "7.2" ]
         packages:
-          # All versions below should be test on PHP ^7.1 (Sentry SDK requirement)
+          # All versions below should be test on PHP ^7.1 (Sentry SDK requirement) and PHP < 8.0 (PHPUnit requirement)
           - { laravel: ^6.0,  testbench: 4.7.*, phpunit: 8.4.* }
           - { laravel: ^7.0,  testbench: 5.1.*, phpunit: 8.4.* }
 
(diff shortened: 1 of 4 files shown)


cleptric:
You can have a look how this was solved in the PHP SDK here https://github.com/getsentry/sentry-php/blob/master/src/Monolog/CompatibilityProcessingHandlerTrait.php

jnoordsij:
Also note that in https://github.com/jnoordsij/sentry-laravel/pull/2 I also added some package combos with Laravel 6/7 and PHP 8.0 that seem to be working; you may want to add those to the tests as well if you intend to support those.

stayallive:
This looks good to me so far, we'll have to see if we need to make any more changes on an actual 10.x install a little later down the line. The only actual thing missing is a changelog entry 👍 

I don't think we need to drop support for anything, we try to support as much as possible without it being a total drag, but with the monolog thing solved I think we can leave the support as it is.

jnoordsij:
Thanks for reviewing! I think the Laravel 10 API should be more or less stable from here, as I've seen them adding and merging Laravel 10 support to many of their self-maintained packages. But I can't really find anything promising this, so I understand if you want to hold this back for a little longer.

Alternatively, is it possible to merge this into a (new) feature branch on this repo, e.g. `feature/l10`, like some other packages are doing? This allows people (and other packages) to add a `dev-branchname` dependency in their `composer.json` to test and prepare for the new release more easily.

stayallive:
Thanks again for working on this 💪

PR #633, opened 2023-01-13
Fix missing breadcrumbs for jobs throwing an exception

### LINKED ISSUE #632: [Laravel Horizion] Breadcrumbs in Job handle function do not get recorded.
### Environment

Laravel: v9.44.0
Sentry: 3.12.0
Laravel-Sentry: 3.1.2
Sentry-SDK: 3.3.0

### Steps to Reproduce

Create a breadcrumb inside a job’s `handle` function.
```
\Sentry\Laravel\Integration::addBreadcrumb(new \Sentry\Breadcrumb(
    $level,
    $type,
    $category,
    $message,
    $data
));
```

Throw an exception inside the job’s `handle` function.

### Expected Result

Being able to see the breadcrumb inside the issue on Sentry.

### Actual Result

The breadcrumb does not appear.

stayallive:
Fixes #632.

There is a downside with this "fix" (hence still a draft).

We might be leaking memory here since we never cleanup any breadcrumbs from failed queue jobs. When a queue workers is running this might be an issue where these breadcrumbs will never be visible but are kept in memory. So when a queue worker churns through say 500 jobs that threw exceptions it has 1500 breadcrumbs (3 per scope) and 500 scopes (unless we limit the amount of scopes already!?).

So not a 100% sure this is the way.

A fix for this can be where we pop the scope if we need to before we run a job, however that breaks [this](https://github.com/getsentry/sentry-laravel/pull/633/files#diff-212611d9baa058b4d98139c7480ad5b39c8f94a5425e2f6e6f5698815c422489R43-R51) test where the middle breadcrumb from this test is being lost which is not ideal either.

### DIFF
--- a/src/Sentry/Laravel/EventHandler.php
+++ b/src/Sentry/Laravel/EventHandler.php
@@ -376,8 +376,6 @@ class EventHandler
 
     protected function queueJobExceptionOccurredHandler(QueueEvents\JobExceptionOccurred $event): void
     {
-        $this->cleanupScopeForTaskWithinLongRunningProcessWhen($this->pushedQueueScopeCount > 0);
-
         $this->afterTaskWithinLongRunningProcess();
     }
 
--- a/src/Sentry/Laravel/EventHandler.php
+++ b/src/Sentry/Laravel/EventHandler.php
@@ -409,6 +409,8 @@ class EventHandler
 
     protected function queueJobProcessingHandler(QueueEvents\JobProcessing $event): void
     {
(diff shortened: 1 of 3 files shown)


cleptric:
Do we hold references for these scopes/breadcrumbs somewhere? If so, we could purge them after X accumulated.

stayallive:
To prevent issues with scopes / breadcrumbs pilling up we pop the scope (if we had pushed one) before starting a new job / scope. For the mostly happy paths in the tests this works correctly. This seems like an improvement but feel free to poke holes in my assumptions!

PR #639, opened 2023-01-30
Register event handlers in DI container

### LINKED ISSUE #637: [Breadcrumbs] [Logs] add to breadcrumbs only from specific logs channels
Hi! I use few laravel log channels at the same time for different log management systems.

But if I enable `breadcrumbs.logs` in `sentry.php` config — all logs from all channel will attached to sentry breadcrumbs.

But I don't want to see messages from all channels in breadcrumbs.

I would like to be able to configure which channels names will be sent to breadcrumbs (include/exclude). Or include/exclude by channel driver.

https://github.com/laravel/framework/discussions/45830

stayallive:
Fixes #637.

I registered the EventHandler classes with the DI container so it can be overriden and extended. However, we won't have any guarantees about BC. So this should be used "at your own risk".


cleptric (approved):
Love the pragmatism! :shipit:

cleptric:
Holding off on this for the moment, as we got no feedback that this caters to the use-case of the initial issue creator.

PR #640, opened 2023-01-30
feat: Add http client breadcrumbs

cleptric:
This adds breadcrumbs for HTTP client calls.

The whole URI dance is done due to:

1. Laravel does not expose the URI on the `Illuminate\Http\Client\Request` class
2. We need to remove the authority, any query strings as well as the fragment from the breadcrumb `url`, and set the query and fragment on their correlating `http.xxx` key. This is done so Relay can apply proper scrubbing without any guesswork. See https://develop.sentry.dev/sdk/data-handling/#breadcrumbs for more context.

We could also make the argument we likely will need the URI part in all three SDKs, so maybe the generic SDK is a good place for this as well.

### DIFF
--- a/src/Sentry/Laravel/EventHandler.php
+++ b/src/Sentry/Laravel/EventHandler.php
@@ -555,9 +555,7 @@ class EventHandler
         }
 
         $level = Breadcrumb::LEVEL_INFO;
-        if ($event->response->clientError()) {
-            $level = Breadcrumb::LEVEL_WARNING;
-        } elseif ($event->response->serverError()) {
+        if ($event->response->failed()) {
             $level = Breadcrumb::LEVEL_ERROR;
         }
 
--- a/src/Sentry/Laravel/EventHandler.php
+++ b/src/Sentry/Laravel/EventHandler.php
@@ -670,9 +670,9 @@ class EventHandler
 
     /**
      * Construct a full URI
-     * 
-     * @param string $url 
-     * @return UriInterface 
+     *
+     * @param string $url
+     * @return UriInterface
      */
     private function getFullUri(string $url): UriInterface
     {
@@ -681,9 +681,9 @@ class EventHandler
 
     /**
      * Construct a partial URI, excluding the authority and the query and fragment parts.
-     * 
-     * @param UriInterface $uri 
-     * @return UriInterface 
+     *
+     * @param UriInterface $uri
+     * @return UriInterface
      */
     private function getPartialUri(UriInterface $uri): UriInterface
     {
--- a/src/Sentry/Laravel/EventHandler.php
+++ b/src/Sentry/Laravel/EventHandler.php
@@ -3,8 +3,6 @@
 namespace Sentry\Laravel;
 
 use Exception;
-use GuzzleHttp\Psr7\Exception\MalformedUriException;
-use GuzzleHttp\Psr7\Uri;
 use Illuminate\Auth\Events as AuthEvents;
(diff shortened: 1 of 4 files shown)


src/Sentry/Laravel/EventHandler.php
```diff
@@ -534,6 +548,58 @@ private function cleanupScopeForOctane(): void
...
     }
 
+    protected function httpClientResponseReceivedHandler(HttpClientEvents\ResponseReceived $event): void
+    {
+        if (!$this->recordHttpClientRequests) {
+            return;
+        }
+
+        $level = Breadcrumb::LEVEL_INFO;
+        if (!$event->response->successful()) {
```
cleptric, line 558: TBD what we should consider as an error.

src/Sentry/Laravel/EventHandler.php
```diff
@@ -534,6 +548,58 @@ private function cleanupScopeForOctane(): void
...
+            $level = Breadcrumb::LEVEL_ERROR;
+        }
+
+        $fullUri = $this->getFullUri($event->request->url());
+        $partialUri = $this->getPartialUri($fullUri);
```
cleptric, line 563: I couldn't come up with a better way to properly generate both versions we need here.
The API of the `GuzzleHttp\Psr7` class is quite narrow.
I also wanted to avoid any string concatenation that will likely be malformed in the end.

src/Sentry/Laravel/EventHandler.php
```diff
@@ -601,4 +667,31 @@ private function cleanupScopeForTaskWithinLongRunningProcessWhen(bool $when): vo
...
+     */
+    private function getPartialUri(UriInterface $uri): UriInterface
+    {
+        return Uri::fromParts([
+            'scheme' => $uri->getScheme(),
+            'host' => $uri->getHost(),
+            'port' => $uri->getPort(),
+            'path' => $uri->getPath(),
+        ]);
+    }
```
cleptric, line 696: We need to apply the same logic for HTTP client spans as well.
@stayallive What would be a good place to share this logic?

src/Sentry/Laravel/EventHandler.php
stayallive: This seems fine as it is, I'd maybe inline it since the `$partilUri` is only used once, but meh. That's just me, this is fine IMHO!

src/Sentry/Laravel/EventHandler.php
stayallive: Seems like that should come from higher up and be defined in some Sentry document what is considered a successful response and not.

stayallive (commented):
We should really think of a way to combine the tracing and breadcrumb event handlers since we are doing so much duplicated work which is a shame. I don't have an immediate answer on how to (best) achieve that though... but this is already pretty good so I'd be happy to go with this and look at refactoring later.

src/Sentry/Laravel/EventHandler.php
cleptric: Went with `LEVEL_WARNING` for 4XX and `LEVEL_ERROR` for 5XX

src/Sentry/Laravel/EventHandler.php
stayallive: How about a trait for now? And we work on a larger event system refactor in another PR?

cleptric: Sounds great!

stayallive (approved):

PR #641, opened 2023-01-30
ref: Add `tracing.http_client_requests` option

cleptric:
We should allow people to opt out of HTTP client tracing. Feels like we missed this in https://github.com/getsentry/sentry-laravel/pull/585 🙂

### DIFF
--- a/src/Sentry/Laravel/Tracing/EventHandler.php
+++ b/src/Sentry/Laravel/Tracing/EventHandler.php
@@ -80,6 +80,13 @@ class EventHandler
     private $traceQueueJobsAsTransactions;
 
     /**
+     * Indicates if we should trace HTTP client requests.
+     *
+     * @var bool
+     */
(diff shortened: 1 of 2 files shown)


src/Sentry/Laravel/Tracing/EventHandler.php
```diff
@@ -108,6 +115,8 @@ public function __construct(array $config, BacktraceHelper $backtraceHelper)
         $this->traceSqlQueries = ($config['sql_queries'] ?? true) === true;
         $this->traceSqlQueryOrigins = ($config['sql_origin'] ?? true) === true;
 
+        $this->traceHttpClientRequests = ($config['http_client_requests'] ?? true) === true;
```
cleptric, line 118: Opted for enabled by default, to not break for people that already use it without updating their config.

stayallive (approved):
Good one 👌

PR #647, opened 2023-02-08
feat: Install an JavaScript SDK during sentry:publish

### LINKED ISSUE #646: Install an JavaScript SDK during `sentry:publish`
Laravel offers two starter kits, Breeze and Jetstream with optionally bootstrap an application with JavaScript frameworks like Vue.js or React (with Interia.js) or Livewire.

- When installing the Laravel SDK, prompt the user if they also want to install an JavaScript SDK to report frontend errors to Sentry
- Update their `package.json`, and run `npm/yarn/pnpm install`
- Display a simple code snippet users need to add to their `main.{js|ts} file`
- We reuse the same DSN as the Laravel application for now

The focus is on frontend error reporting, performance monitoring might be added in the future.

**Vue.js snippet**

```js
import * as Sentry from "@sentry/vue";
import { BrowserTracing } from "@sentry/tracing";

Sentry.init({
  app,
  dsn: "https://examplePublicKey@o0.ingest.sentry.io/0",
  integrations: [new BrowserTracing()],
});
```

**React snippet**

```js
import * as Sentry from "@sentry/react";
import { BrowserTracing } from "@sentry/tracing";

Sentry.init({
  dsn: "https://examplePublicKey@o0.ingest.sentry.io/0",
  integrations: [new BrowserTracing()],
});
```

**Livewire snippet**

```js
// TBD
```

cleptric:
Screenshot_2023-03-06_at_11 33 34

This is a little PoC to offer the installation of one of our JavaScript SDKs when running `php artisan sentry:publish`.

closes #646

### DIFF
--- a/src/Sentry/Laravel/Console/PublishCommand.php
+++ b/src/Sentry/Laravel/Console/PublishCommand.php
@@ -17,7 +17,10 @@ class PublishCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'sentry:publish {--dsn=} {--without-performance-monitoring} {--without-test}';
+    protected $signature = 'sentry:publish {--dsn=}
+                                           {--without-performance-monitoring}
+                                           {--without-test}
+                                           {--without-javascript-sdk}';
 
     /**
      * The console command description.
@@ -25,6 +28,12 @@ class PublishCommand extends Command
      * @var string
      */
     protected $description = 'Publishes and configures the Sentry config.';
+
+    protected const SDK_CHOICE_BROWSER = 'JavaScript (default)';
+    protected const SDK_CHOICE_VUE     = 'Vue.js';
+    protected const SDK_CHOICE_REACT   = 'React';
+    protected const SDK_CHOICE_ANGULAR = 'Angular';
+    protected const SDK_CHOICE_SVELTE  = 'Svelte';
 
     /**
      * Execute the console command.
@@ -75,13 +84,14 @@ class PublishCommand extends Command
             }
         }
 
-        $this->askForJavaScriptSdk($dsn);
-
         $this->info('Publishing Sentry config...');
         $this->call('vendor:publish', ['--provider' => ServiceProvider::class]);
 
         if (!$this->setEnvValues($env)) {
             return 1;
+        }
+        if ($this->confirm('Do you want to install one of our JavaScript SDKs?', !$this->option('without-javascript-sdk'))) {
+            $this->installJavaScriptSdk($dsn);
         }
 
         return 0;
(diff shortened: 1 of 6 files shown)


stayallive (commented):
Only thing really is maybe to use [`import.meta.env.SENTRY_LARAVEL_DSN`](https://github.com/laravel/laravel/blob/10.x/resources/js/bootstrap.js#LL25C13-L25C29) instead of hardcoding the DSN, otherwise this looks like it's more than sufficient!

This might be good enough because you would only use this in new packages, but another options would be to add a `// @TODO: Read DSN from environment instead of hardcoded` or something?

cleptric:
> [import.meta.env.SENTRY_LARAVEL_DSN](https://github.com/laravel/laravel/blob/10.x/resources/js/bootstrap.js#LL25C13-L25C29)

I think we need to prefix them with `VITE_`.

> To prevent accidentally leaking env variables to the client, only variables prefixed with VITE_ are exposed to your Vite-processed code. e.g. for the following env variables:

https://vitejs.dev/guide/env-and-mode.html#env-files

Let's target Laravel 10/9 with Vite, as 8 and below use Webpack. I don't see many people creating new projects with these older versions.

I'll also add another "Vanilla.js" choice to install our default browser integration.

Additionally, we can also expand the snippets, as all Laravel version ship with Axios by default, so we can also add performance monitoring.

```
import * as Sentry from "@sentry/browser";
import { BrowserTracing } from "@sentry/tracing";

Sentry.init({
  dsn: import.meta.env.VITE_APP_SENTRY_LARAVEL_DSN,
  integrations: [new BrowserTracing()],
  tracesSampleRate: 1.0,
});
```

Brb!

stayallive:
Ah right, the prefix, yes! We also need to publish that env variable then in addition to the regular DSN env variable, we can just add `VITE_SENTRY_LARAVEL_DSN="${SENTRY_LARAVEL_DSN}"` though to expose it with an alias instead of duplicating it.

src/Sentry/Laravel/Console/PublishCommand.php
```diff
@@ -151,4 +155,115 @@ private function askForDsnInput(): string
...
+            $snippet = '';
+
+            switch ($framework) {
+                case 'Vue.js':
+                    $this->updateNodePackages(function ($packages) {
+                        return [
+                            '@sentry/vue' => '^7.36.0',
+                        ] + $packages;
+                    });
+                    $snippet = 'import * as Sentry from "@sentry/vue";' . PHP_EOL . PHP_EOL;
```
AbhiPrasad, line 176: m: Can we keep these snippets in js files we read from file system? This way we can also validate if syntax is valid (also makes diff's easier to read in the future).

cleptric: Sure can do, it's also less of a string concatenation mess 😬

src/Sentry/Laravel/Console/PublishCommand.php
```diff
@@ -151,4 +165,121 @@ private function askForDsnInput(): string
...
+                $this->updateNodePackages(function ($packages) {
+                    return [
+                        '@sentry/svelte' => '^7.40.0',
+                    ] + $packages;
+                });
+                $snippet = file_get_contents(__DIR__ . '/../../../../stubs/sentry-javascript/svelte.js');
+                break;
+        }
+
+        $env['VITE_SENTRY_DSN_PUBLIC'] ='"${SENTRY_LARAVEL_DSN}"';
```
HazAT, line 228: I would rename this `VITE_SENTRY_DSN_PUBLIC` everywhere to `SENTRY_LARAVEL_DSN`

HazAT (changes requested):

src/Sentry/Laravel/Console/PublishCommand.php
cleptric: See https://github.com/getsentry/sentry-laravel/pull/647#issuecomment-1434576854, there is a technical reason (the best reason) why we need to prefix your env vars with `VITE_`

HazAT: cool 👍

HazAT (approved):

stayallive (approved):

PR #651, opened 2023-02-14
Make the Sentry handler safer dealing with unknown values

### LINKED ISSUE #644: Adding a string with the key `user` to loging context throws an exception
### Environment

Sentry: 3.1.3
Laravel: 9.43.0

### Steps to Reproduce

1. Use sentry as the logger driver

```
        'stack' => [
            'driver' => 'stack',
            'channels' => ['daily', 'sentry'],
            'ignore_exceptions' => false,
        ],
```

2. Send a log

```
logger()->info('user', ['userId' => 5, 'user' => 'abc'])
```

### Expected Result

The log is added without any problems.

If this is expected behaviour, this should be explicitly documented somewhere

### Actual Result

An exception is thrown, b/c `user` is apparently a reserved keyword, and passing it in the log creates an exception

image

stayallive:
I've added a bunch of checks and try to err on the side of caution where we might not set values on the scope but try to always get the event out the door instead of crashing.

Also added a few tests, long way from exhaustive but better than before.

Fixes #644.

### DIFF
--- a/src/Sentry/Laravel/SentryHandler.php
+++ b/src/Sentry/Laravel/SentryHandler.php
@@ -2,6 +2,7 @@
 
 namespace Sentry\Laravel;
 
+use DateTimeInterface;
 use Monolog\DateTimeImmutable;
 use Monolog\Logger;
 use Monolog\LogRecord;
@@ -15,6 +16,7 @@ use Sentry\Severity;
 use Sentry\State\HubInterface;
 use Sentry\State\Scope;
 use Throwable;
+use TypeError;
 
 class SentryHandler extends AbstractProcessingHandler
 {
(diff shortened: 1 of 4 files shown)


src/Sentry/Laravel/SentryHandler.php
```diff
@@ -301,4 +291,65 @@ public function addBreadcrumb(Breadcrumb $crumb): self
...
+                unset($context['user']);
+            } catch (TypeError $e) {
+                // In some cases the context can be invalid, in that case we ignore it and
+                // choose to not send it to Sentry in favor of not breaking the application
+            }
+        }
+    }
+
+    /**
+     * Test if the value passed can be casted to a string.
```
cleptric, line 345:
```suggestion
     * Check if the value passed can be cast to a string.
```

cleptric (approved):

PR #656, opened 2023-02-27
Cache performance spans & breadcrumbs

### LINKED ISSUE #627: Cache performance spans
With the `Illuminate\Redis\Events\CommandExecuted` we could add spans for redis commands. We do this in the Symfony SDK where cache calls are profiled, maybe we could generelize since just Redis spans don't seem too interesting but cache performance spans could ne nice to see since they do take time but haven't looked to close if there are command where we can hook into the cache events.

stayallive:
This adds:

- Cache breadcrumbs
- Redis command spans (optional)

We might want to extend this in the future to also trace the cache commands themselfs (like we do in Symfony) which requires decorating the cache drivers (not enough events to do it purely event based).

Fixes #627

Requires merging #657 first!

### DIFF
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -15,6 +15,9 @@ return [
         // Capture Laravel logs in breadcrumbs
         'logs' => true,
 
+        // Capture Laravel cache events in breadcrumbs
+        'cache' => true,
+
         // Capture Livewire components in breadcrumbs
         'livewire' => true,
 
@@ -55,6 +58,12 @@ return [
 
         // Capture HTTP client requests as spans
         'http_client_requests' => true,
+
+        // Capture Redis operations as spans (this enables Redis events in Laravel)
+        'redis_commands' => env('SENTRY_TRACE_REDIS_COMMANDS', false),
+
+        // Try to find out where the Redis command originated from and add it to the command spans
+        'redis_origin' => true,
 
         // Indicates if the tracing integrations supplied by Sentry should be loaded
         'default_integrations' => true,
--- a/src/Sentry/Laravel/Tracing/EventHandler.php
+++ b/src/Sentry/Laravel/Tracing/EventHandler.php
@@ -233,7 +233,7 @@ class EventHandler
             $queryOrigin = $this->resolveQueryOriginFromBacktrace();
 
             if ($queryOrigin !== null) {
-                $context->setData(['sql.origin' => $queryOrigin]);
+                $context->setData(['db.sql.origin' => $queryOrigin]);
             }
         }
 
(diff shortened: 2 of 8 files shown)


src/Sentry/Laravel/Tracing/EventHandler.php
```diff
@@ -255,6 +275,45 @@ private function resolveQueryOriginFromBacktrace(): ?string
...
+        // If there is no tracing span active there is no need to handle the event
+        if ($parentSpan === null) {
+            return;
+        }
+
+        // @TODO: Do we need/want to prefix all data with `db.redis`? It's already in the `db.redis` op so maybe it's not needed?
+        $data = [
+            'db.redis.connection' => $event->connectionName,
+            // @TODO: Sending all the parameters should possibly be locked behind `send_default_pii`? It does contain sensitive data but also Redis integration is opt-in so maybe it's fine?
+            'db.redis.parameters' => $event->parameters,
```
stayallive, line 295: The Ruby SDK locks this behind `send_default_pii` which makes sense... what is common sense here?

src/Sentry/Laravel/Tracing/EventHandler.php
```diff
@@ -255,6 +275,45 @@ private function resolveQueryOriginFromBacktrace(): ?string
...
+        }
+
+        $parentSpan = SentrySdk::getCurrentHub()->getSpan();
+
+        // If there is no tracing span active there is no need to handle the event
+        if ($parentSpan === null) {
+            return;
+        }
+
+        // @TODO: Do we need/want to prefix all data with `db.redis`? It's already in the `db.redis` op so maybe it's not needed?
```
stayallive, line 291: Can't find much guidance on what to do here exactly, do we prefix all data with the OP or no (or sometimes?)?

config/sentry.php
```diff
@@ -44,6 +47,12 @@
         // Capture views as spans
         'views' => true,
 
+        // Capture Redis operations as spans (this enables Redis events in Laravel)
+        'redis_commands' => env('SENTRY_TRACE_REDIS_COMMANDS', false),
```
stayallive, line 51: Disabled by default since it enables events on the Redis connections (which are disabled by default) but maybe there is no reason to be THAT safe and we can just enable it (and allow users to disable it if they don't want it)? We could also enable it by default and have the user do `Redis::enableEvents()` in their code to fully enable the feature.

src/Sentry/Laravel/Tracing/EventHandler.php
cleptric: Let's gate it behind `send_default_pii` and/or make it configurable as with SQL params.

src/Sentry/Laravel/Tracing/EventHandler.php
cleptric: You can keep it as is.

cleptric (approved):

PR #657, opened 2023-03-02
Add Livewire tracing integration

stayallive:
This renames transaction names to `livewire?component={component-name}` if a Livewire message request is running (message requests are how Livewire syncs state between frontend and backend). The benefit here is that otherwise _all_ message requests are grouped as `/livewire/message/{name}` which is not very useful at all.

Also added spans when Livewire components are being rendered in both the message and normal requests.

And added breadcrumbs, the `component.mount` breadcrumb with the data for "normal" page requests and a `component.booted` breadcrumb for those message requests with the `update` info attached so you can see what update is going to be made to the component. These breadcrumbs are mostly interesting for exceptions so you can see what data / mutations are running on the components.

image

image

This is a quick PoC and will be refined some more in the coming days, any feedback is welcome!

### DIFF
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -15,6 +15,9 @@ return [
         // Capture Laravel logs in breadcrumbs
         'logs' => true,
 
+        // Capture Livewire components in breadcrumbs
+        'livewire' => true,
+
         // Capture SQL queries in breadcrumbs
         'sql_queries' => true,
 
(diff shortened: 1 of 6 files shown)


SebastiaanKloos:
Yes this is very helpful! ❤️

cleptric:
Mr. Span Ops himself, @AbhiPrasad, rightfully pointed out to use a more aligned span op like `ui.livewire.component`. We can add this to https://develop.sentry.dev/sdk/performance/span-operations/

cleptric (approved):

PR #658, opened 2023-03-05
Fix incorrect annotation on Facade

leo108:
`captureEvent` should accept `\Sentry\Event` instead of `\Throwable`


cleptric (approved):
Thank you

neodc:
Hello, would it be possible to get a tagged release with this fix? Thanks

PR #664, opened 2023-03-19
Fix scheduled commands running in background not reporting success/failure

miskith:
Fix for scheduled events running in background to correctly report it's state


cleptric:
Could you expand on what this fixes exactly?

miskith:
> Could you expand on what this fixes exactly?

When running scheduled task in background ( https://laravel.com/docs/10.x/scheduling#background-tasks ), success or fail statuses are not reported and cron runs in report are just timing out.

stayallive:
I've refactored your PR a bit to use a cache entry per command instead (and also make them expire at the same time as a mutex does) since your idea of putting all the cache entries in a single array in the cache is subject to race conditions which is less ideal and also might end up only increasing in size if for some reasons some commands do not finish properly.

cleptric (approved):

PR #668, opened 2023-03-22
Validate first cache command argument type before using as string

### LINKED ISSUE #663: Object of class Closure could not be converted to string
### Environment

`/vendor/sentry/sentry-laravel/src/Sentry/Laravel/Features/CacheIntegration.php:84`
```php
84         $context->setDescription(strtoupper($event->command) . ' ' . ($event->parameters[0] ?? null));
```

### Steps to Reproduce

1. nothing, it just happens for anything uses cache.

### Expected Result

everything to work

### Actual Result

Object of class Closure could not be converted to string

stayallive:
Horizon uses pipelining and the first argument is a closure instead the assumed string.

We first validate the assumption before using the first Redis command argument.

Also a check for newlines since scripts being executed do not look nice nor are useful in the description.

Fixes #663

### DIFF
--- a/src/Sentry/Laravel/Features/CacheIntegration.php
+++ b/src/Sentry/Laravel/Features/CacheIntegration.php
@@ -6,6 +6,7 @@ use Illuminate\Cache\Events;
 use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Redis\Events as RedisEvents;
 use Illuminate\Redis\RedisManager;
+use Illuminate\Support\Str;
 use Sentry\Breadcrumb;
 use Sentry\Laravel\Features\Concerns\ResolvesEventOrigin;
 use Sentry\Laravel\Integration;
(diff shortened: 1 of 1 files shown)


src/Sentry/Laravel/Features/CacheIntegration.php
```diff
@@ -81,7 +82,16 @@ public function handleRedisCommand(RedisEvents\CommandExecuted $event): void
...
+
+        $keyForDescription = '';
+
+        // If the first parameter is a string and does not contain a newline we use it as the description since it's most likely a key
+        // This is not a perfect solution but it's the best we can do without understanding the command that was executed
+        if (!empty($event->parameters[0]) && is_string($event->parameters[0]) && !Str::contains($event->parameters[0], "\n")) {
+            $keyForDescription = $event->parameters[0];
+        }
+
+        $context->setDescription(rtrim(strtoupper($event->command) . ' ' . $keyForDescription));
```
cleptric, line 94: Might add a blank ` ` at the end, but not a biggie.

cleptric (approved):

src/Sentry/Laravel/Features/CacheIntegration.php
stayallive: Have an `rtrim` to prevent that 😉

PR #680, opened 2023-04-18
fix: Fix CheckIn constructor argument order

cleptric:
The signature is

```php
    public function __construct(
        string $monitorSlug,
        CheckInStatus $status,
        string $id = null,
        ?string $release = null,
        ?string $environment = null,
        $duration = null
    ) {
```


stayallive (approved):
🤦‍♂️

Issue #684, opened 2023-04-26
Cannot install on Lumen 10.x 


szwagros:
When trying to install sentry on Lumen 10.x (running composer require "sentry/sentry-laravel:^2.14") I get this error:

  Problem 1
    - Root composer.json requires sentry/sentry-laravel ^2.14 -> satisfiable by sentry/sentry-laravel[2.14.0, 2.14.1, 2.14.2].
    - sentry/sentry-laravel[2.14.0, ..., 2.14.2] require illuminate/support 5.0 - 5.8 | ^6.0 | ^7.0 | ^8.0 | ^9.0 -> found illuminate/support[v5.0.0, ..., v5.8.36, v6.0.0, ..., v6.20.44, v7.0.0, ..., v7.30.6, v8.0.0, ..., v8.83.27, v9.0.0, ..., v9.52.7] but the package is fixed to v10.2.0 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.

Is Lumen 10 version not supported ?

stayallive:
Hi @szwagros, we indeed did not take Lumen support with us with the Laravel 10 push. We are trying to fix that and have started work on adding it back in, if you are willing you can test it out by installing the lumen branch: #685. If you find any issues be sure to let us know so we can fix that up before we release this!

stayallive:
Hey Tim, much appreciated! Thanks for letting us know 😄

szwagros:
> Hi @szwagros, we indeed did not take Lumen support with us with the Laravel 10 push. We are trying to fix that and have started work on adding it back in, if you are willing you can test it out by installing the lumen branch: #685. If you find any issues be sure to let us know so we can fix that up before we release this!

Hi - sorry for late replay. Yes it installs and works fine (error reporting, adding scope etc.) I did not test perf monitoring etc.

stayallive:
Sorry this was closed a few minuted too early but [`3.4.0`](https://github.com/getsentry/sentry-laravel/releases/tag/3.4.0) was just released and contains Lumen 10.x support!

Thanks all for testing!

PR #685, opened 2023-05-01
Reintroduce Lumen support

stayallive:
Lumen was dropped in #579, but this might have been a little too soon.

So, here it is, back again with support for Laravel 10.x components!

Also a little improved where we should see better transaction names more often hopefully.

This was tested quite quickly so a little more testing would be helpful, you can install this by running the following:

```
composer require "sentry/sentry-laravel:dev-lumen@dev"
```

After this you can continue with the docs to register Sentry in your `bootstrap/app.php`: https://docs.sentry.io/platforms/php/guides/laravel/other-versions/lumen/

Relates: #684.


README.md
```diff
@@ -97,7 +97,6 @@ The Laravel and Lumen version listed below were supported in previous versions:
 - Laravel `<= 4.2.x` is supported until `0.8.x`
 - Laravel `<= 5.7.x` on PHP `<= 7.0` is supported until `0.11.x`
 - Laravel `>= 5.x.x` on PHP `>= 7.1` is supported until `2.14.x`
-- Laravel Lumen is supported until `2.14.x`
```
cleptric, line 100: We should rephrase
> The Laravel and Lumen version listed below were supported in previous versions:

cleptric (approved):
Besides the minor update in the README, this looks great!

```php
$router->get('example[/{id}]', 'ExampleController@show');
```

Screenshot 2023-05-08 at 04 22 28

PR #687, opened 2023-05-03
Replace sentry/sdk by sentry/sentry 3.17

nicolas-grekas:
Same as #649 , but updated to 3.17 thanks to https://github.com/getsentry/sentry-php/pull/1504


cleptric:
Holding off on this for the moment. We don't need to fix what is not broken.
We'll revisit this in the next major of the base SDK.

PR #698, opened 2023-05-22
Allow setting up features in an inactive state

### LINKED ISSUE #696: Cron monitoring without DSN set results in error
### Discussed in https://github.com/getsentry/sentry-laravel/discussions/695

<div type='discussions-op-text'>

<sup>Originally posted by **lloricode** May 20, 2023</sup>
I have a scheduled command

```php

        $schedule->command(SomeCommand::class)
            ->sentryMonitor('point-of-sale-cron-monitoring')
```

When I try to run ` php artisan schedule:list` on my local, it's working fine, but when `SENTRY_LARAVEL_DSN=` in .env is not set it will throw an exception:

```
Method Illuminate\Console\Scheduling\Event::sentryMonitor does not exist.
```

I hope this allow to just ignore when there is no setup for sentry</div>

stayallive:
Some features might create macro's that should be defined if the SDK is active or not so Ive introduced a new setup method for features so they can register inert versions of their macro's. It isn't a nice approach but I figured it's the nicest way to prevent doing too much work or needing to do much guarding whenever a DSN is not set,

This fixes #696.

### DIFF
--- a/src/Sentry/Laravel/Features/ConsoleIntegration.php
+++ b/src/Sentry/Laravel/Features/ConsoleIntegration.php
@@ -56,6 +56,13 @@ class ConsoleIntegration extends Feature
         });
     }
 
+    public function setupInactive(): void
+    {
+        SchedulingEvent::macro('sentryMonitor', function (string $monitorSlug) {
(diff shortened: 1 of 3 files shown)


cleptric (approved):
:shipit:

PR #719, opened 2023-06-26
Tracing without Performance

### LINKED ISSUE #713: Tracing without Performance
This will add Tracing without Performance to the Laravel SDK, a follow-up of https://github.com/getsentry/sentry-php/issues/1525.

cleptric:
Closes #713

### DIFF
--- a/src/Sentry/Laravel/Integration.php
+++ b/src/Sentry/Laravel/Integration.php
@@ -13,12 +13,15 @@ use Sentry\SentrySdk;
 use Sentry\Severity;
 use Sentry\Tracing\TransactionSource;
 use Throwable;
-use function Sentry\addBreadcrumb;
-use function Sentry\configureScope;
 use Sentry\Breadcrumb;
 use Sentry\Event;
 use Sentry\Integration\IntegrationInterface;
 use Sentry\State\Scope;
+
+use function Sentry\addBreadcrumb;
+use function Sentry\configureScope;
+use function Sentry\getBaggage;
+use function Sentry\getTraceparent;
 
 class Integration implements IntegrationInterface
 {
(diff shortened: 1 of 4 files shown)


src/Sentry/Laravel/Integration.php
```diff
@@ -192,13 +189,7 @@ public static function sentryTracingMeta(): string
...
     public static function sentryBaggageMeta(): string
     {
-        $span = SentrySdk::getCurrentHub()->getSpan();
-
-        if ($span === null) {
-            return '';
-        }
-
-        return sprintf('<meta name="baggage" content="%s"/>', $span->toBaggage());
+        return sprintf('<meta name="baggage" content="%s"/>', getBaggage());
```
cleptric, line 192: As we'll very likely always get some value here, I think it's OK to omit checks here.

cleptric:
@stayallive the last TODO here would be to check where we would need to create a new propagation context, especially in a queue context. Today, we control this via `queue_job_transactions`, so maybe we can move this flag into `queue_job_traces` or something.

stayallive (approved):
This looks 🍑-y to me!

cleptric:
The propagation context does not concern us that much in the case of transactions. It's more about which `traceId` is used for error events. 

I opted to attach a new `PropagationContext` to the span once `pushScope()` is called, hence linking all errors that happen in this specific queue job together.
In case the payload contains a traceparent and baggage payload, and the SDK is configured for `queue_job_transactions`, the `PropagationContext` on the scope will be overwritten by `continueTrace()`, hence continuing the trace with the correct `traceId`.

stayallive:
Good call, that should be fine!

stayallive (approved):

PR #720, opened 2023-06-26
GraphQL and HTTP client improvements

stayallive:
- Set `api_target` on GraphQL request bodies to `graphql` to allow Sentry to highlight it nicely

    image

    Since we are not technically doing a HTTP client request but we are the GraphQL server I have not included the response since we have never done that in PHP yet. And since I couldn't find any docs on the GraphQL server thingies I tried this (2d35c899334888fed0190a50d4c39736751aa076) and it looks very nice but not a 100% confident it is "correct".

- Add missing `response_body_size` and `request_body_size` to HTTP client breadcrumbs (https://develop.sentry.dev/sdk/features/#http-client-integrations)

### DIFF
--- a/src/Sentry/Laravel/EventHandler.php
+++ b/src/Sentry/Laravel/EventHandler.php
@@ -291,6 +291,8 @@ class EventHandler
                 'status_code' => $event->response->status(),
                 'http.query' => $fullUri->getQuery(),
                 'http.fragment' => $fullUri->getFragment(),
+                'request_body_size' => strlen($event->request->body()),
+                'response_body_size' => strlen($event->response->body()),
             ]
         ));
     }
@@ -313,6 +315,7 @@ class EventHandler
                 'method' => $event->request->method(),
                 'http.query' => $fullUri->getQuery(),
                 'http.fragment' => $fullUri->getFragment(),
+                'request_body_size' => strlen($event->request->body()),
             ]
         ));
     }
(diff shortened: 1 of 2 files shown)


cleptric (approved):

cleptric:
@stayallive Want to move this forward? We can follow up on response handling and errors in another PR.

stayallive:
I'm pretty happy with this 👍

src/Sentry/Laravel/EventHandler.php
```diff
@@ -291,6 +291,8 @@ protected function httpClientResponseReceivedHandler(HttpClientEvents\ResponseRe
                 'status_code' => $event->response->status(),
                 'http.query' => $fullUri->getQuery(),
                 'http.fragment' => $fullUri->getFragment(),
+                'request_body_size' => strlen($event->request->body()),
+                'response_body_size' => strlen($event->response->body()),
```
marandaneto, line 295: request/response body size is in bytes instead of string length.

stayallive: See: https://www.php.net/manual/en/function.strlen.php#refsect1-function.strlen-notes

marandaneto: Yep, my bad, I just assumed because its called `strlen`, I know nothing about PHP :D

stayallive: Easy mistake to make, welcome in PHP land 😉

smeubank:
do we need docs changes for PHP for this?

cleptric:
As we talked about during our daily, there are no changes needed in a user's application to enable the improved GrapQL request payload highlighting.
[GraphQL errors](https://blog.sentry.io/handling-graphql-errors-using-sentry/) are not part of this PR. This might come as a follow-up.

PR #726, opened 2023-07-04
Add tracing for storage access

spawnia:
This pull request adds tracing for storage access.

As usual, the integration is enabled through the configuration.

### DIFF
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -55,6 +55,8 @@ return [
         'views' => true,
 
         // Capture storage access as spans
+        // Filesystems have to opt-in by changing their configuration like this:
+        // ['driver' => 'local', ...] -> ['driver' => 'sentry', 'original_driver' => 'local', ...]
         'storage' => true,
 
         // Capture Livewire components as spans
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -55,8 +55,6 @@ return [
         'views' => true,
 
         // Capture storage access as spans
-        // Filesystems have to opt-in by changing their configuration like this:
-        // ['driver' => 'local', ...] -> ['driver' => 'sentry', 'original_driver' => 'local', ...]
         'storage' => true,
 
         // Capture Livewire components as spans
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -19,6 +19,9 @@ return [
         // Capture Laravel cache events in breadcrumbs
         'cache' => true,
 
+        // Capture storage access as breadcrumbs
+        'storage' => true,
+
         // Capture Livewire components in breadcrumbs
         'livewire' => true,
 
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -54,6 +54,9 @@ return [
         // Capture views as spans
         'views' => true,
 
+        // Capture storage access as spans
+        'storage' => true,
+
         // Capture Livewire components as spans
         'livewire' => true,
 
--- a/src/Sentry/Laravel/EventHandler.php
+++ b/src/Sentry/Laravel/EventHandler.php
@@ -200,7 +200,7 @@ class EventHandler
         }
 
         try {
-            call_user_func_array([$this, $handlerMethod], $arguments);
+            ($this->handlerMethod)(...$arguments);
         } catch (Exception $exception) {
             // Ignore
         }
--- a/src/Sentry/Laravel/EventHandler.php
+++ b/src/Sentry/Laravel/EventHandler.php
@@ -200,7 +200,7 @@ class EventHandler
         }
 
         try {
-            ($this->handlerMethod)(...$arguments);
+            $this->{$handlerMethod}(...$arguments);
         } catch (Exception $exception) {
             // Ignore
         }
--- a/src/Sentry/Laravel/Features/StorageIntegration.php
+++ b/src/Sentry/Laravel/Features/StorageIntegration.php
@@ -20,6 +20,7 @@ class StorageIntegration extends Feature
             $filesystemManager->extend('sentry', function (Application $application, array $config) use ($filesystemManager): Filesystem {
                 $config['driver'] = $config['original_driver'];
                 unset($config['original_driver']);
+
                 $originalFilesystem = $filesystemManager->build($config);
 
                 return new TracingFilesystem($originalFilesystem);
--- a/src/Sentry/Laravel/Features/StorageIntegration.php
+++ b/src/Sentry/Laravel/Features/StorageIntegration.php
@@ -16,6 +16,11 @@ class StorageIntegration extends Feature
 
     public function setup(): void
     {
+        foreach (config('filesystems.disks') as $disk => $config) {
+            config(["filesystems.disks.{$disk}.original_driver" => $config['driver']]);
+            config(["filesystems.disks.{$disk}.driver" => 'sentry']);
+        }
+
         $this->container()->afterResolving(FilesystemManager::class, static function (FilesystemManager $filesystemManager): void {
             $filesystemManager->extend('sentry', function (Application $application, array $config) use ($filesystemManager): Filesystem {
                 $config['driver'] = $config['original_driver'];
@@ -26,10 +31,5 @@ class StorageIntegration extends Feature
                 return new TracingFilesystem($originalFilesystem);
             });
         });
-    }
-
-    public function setupInactive(): void
-    {
-        $this->setup();
     }
 }
--- a/src/Sentry/Laravel/Features/StorageIntegration.php
+++ b/src/Sentry/Laravel/Features/StorageIntegration.php
@@ -65,9 +65,11 @@ class StorageIntegration extends Feature
 
                     $originalFilesystem = $diskResolver($disk, $config);
 
+                    $defaultData = ['disk' => $disk, 'driver' => $config['driver']];
+
                     return $originalFilesystem instanceof CloudFilesystem
-                        ? new TracingCloudFilesystem($originalFilesystem, $disk, $config['driver'])
-                        : new TracingFilesystem($originalFilesystem, $disk, $config['driver']);
+                        ? new TracingCloudFilesystem($originalFilesystem, $defaultData)
+                        : new TracingFilesystem($originalFilesystem, $defaultData);
                 }
             );
         });
--- a/src/Sentry/Laravel/Features/StorageIntegration.php
+++ b/src/Sentry/Laravel/Features/StorageIntegration.php
@@ -2,9 +2,11 @@
 
 namespace Sentry\Laravel\Features;
 
+use Illuminate\Contracts\Filesystem\Cloud as CloudFilesystem;
 use Illuminate\Contracts\Filesystem\Filesystem;
 use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Filesystem\FilesystemManager;
+use Sentry\Laravel\Tracing\Storage\TracingCloudFilesystem;
 use Sentry\Laravel\Tracing\Storage\TracingFilesystem;
 
 class StorageIntegration extends Feature
@@ -28,7 +30,9 @@ class StorageIntegration extends Feature
 
                 $originalFilesystem = $filesystemManager->build($config);
 
-                return new TracingFilesystem($originalFilesystem);
+                return $originalFilesystem instanceof CloudFilesystem
+                    ? new TracingCloudFilesystem($originalFilesystem)
+                    : new TracingFilesystem($originalFilesystem);
             });
         });
     }
--- a/src/Sentry/Laravel/Features/StorageIntegration.php
+++ b/src/Sentry/Laravel/Features/StorageIntegration.php
@@ -2,7 +2,6 @@
 
 namespace Sentry\Laravel\Features;
 
-use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Contracts\Filesystem\Filesystem;
 use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Filesystem\FilesystemManager;
@@ -15,18 +14,21 @@ class StorageIntegration extends Feature
         return $this->isTracingFeatureEnabled('storage');
     }
 
-    public function setup(Dispatcher $events): void
+    public function setup(): void
     {
-        if ($this->isTracingFeatureEnabled('storage', false)) {
-            $this->container()->afterResolving(FilesystemManager::class, static function (FilesystemManager $filesystemManager): void {
-                $filesystemManager->extend('sentry', function (Application $application, array $config) use ($filesystemManager): Filesystem {
-                    $config['driver'] = $config['original_driver'];
-                    unset($config['original_driver']);
-                    $originalFilesystem = $filesystemManager->build($config);
+        $this->container()->afterResolving(FilesystemManager::class, static function (FilesystemManager $filesystemManager): void {
+            $filesystemManager->extend('sentry', function (Application $application, array $config) use ($filesystemManager): Filesystem {
+                $config['driver'] = $config['original_driver'];
+                unset($config['original_driver']);
+                $originalFilesystem = $filesystemManager->build($config);
 
-                    return new TracingFilesystem($originalFilesystem);
-                });
+                return new TracingFilesystem($originalFilesystem);
             });
-        }
+        });
+    }
+
+    public function setupInactive(): void
+    {
+        $this->setup();
     }
 }
--- a/src/Sentry/Laravel/Features/StorageIntegration.php
+++ b/src/Sentry/Laravel/Features/StorageIntegration.php
@@ -2,20 +2,11 @@
 
 namespace Sentry\Laravel\Features;
 
-use Illuminate\Cache\Events;
 use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Contracts\Filesystem\Filesystem;
 use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Filesystem\FilesystemManager;
-use Illuminate\Redis\Events as RedisEvents;
-use Illuminate\Redis\RedisManager;
-use Illuminate\Support\Str;
-use Sentry\Breadcrumb;
-use Sentry\Laravel\Features\Concerns\ResolvesEventOrigin;
-use Sentry\Laravel\Integration;
 use Sentry\Laravel\Tracing\Storage\TracingFilesystem;
-use Sentry\SentrySdk;
-use Sentry\Tracing\SpanContext;
 
 class StorageIntegration extends Feature
 {
--- a/src/Sentry/Laravel/Features/StorageIntegration.php
+++ b/src/Sentry/Laravel/Features/StorageIntegration.php
@@ -6,11 +6,14 @@ use Illuminate\Contracts\Filesystem\Cloud as CloudFilesystem;
 use Illuminate\Contracts\Filesystem\Filesystem;
 use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Filesystem\FilesystemManager;
+use RuntimeException;
 use Sentry\Laravel\Tracing\Storage\TracingCloudFilesystem;
 use Sentry\Laravel\Tracing\Storage\TracingFilesystem;
 
 class StorageIntegration extends Feature
 {
+    private const STORAGE_DRIVER_NAME = 'sentry';
+
     public function isApplicable(): bool
     {
         return $this->isTracingFeatureEnabled('storage');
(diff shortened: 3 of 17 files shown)


src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
```diff
@@ -0,0 +1,153 @@
...
+        $context->setData($data);
+
+        return trace($originalCall, $context);
+    }
+
+    public function assertExists($path, $content = null)
+    {
+        return $this->withTracing('assertExists', ['path' => $path], function () use ($path, $content) {
+            // TODO assertExists only present on FilesystemAdapter
+            return $this->filesystem->assertExists($path, $content);
```
spawnia, line 39: Should we require `FilesystemAdapter` to be used so this always works? Alternatively, we could throw if `! $this->filesystem instanceof FilesystemAdapter` in those particular calls.

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
```diff
@@ -0,0 +1,153 @@
...
+        });
+    }
+
```
spawnia, line 42: TODO add additional methods from `FilesystemAdapter` that are not part of `Illuminate\Contracts\Filesystem\Cloud`.

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
```diff
@@ -0,0 +1,153 @@
...
+    public function readStream($path)
+    {
+        // TODO: Implement readStream() method.
+    }
+
+    public function put($path, $contents, $options = [])
+    {
+        return $this->withTracing('put', ['path' => $path], function () use ($path, $contents, $options) {
```
spawnia, line 67: I thought we better exclude `$contents` from this, but I am not sure. What if it is a very large file, does the SDK handle that?

We could add `$options` though.

If we decide to just add all parameters in any case, the implementation can be significantly simplified.

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
```diff
@@ -0,0 +1,153 @@
...
+     * @template TResult
+     * @param array<string, mixed> $data
+     * @param callable(): TResult $originalCall
+     * @return TResult
+     */
+    protected function withTracing(string $method, array $data, callable $originalCall)
+    {
+        $context = new SpanContext();
+        $context->setOp("storage.{$method}");
+        $context->setData($data);
```
spawnia, line 30: Should we add a custom `$description` for a better user experience, or is just data sufficient?

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
cleptric: This has the potential to increase the payload size significantly, which might lead to dropped events that are hard to debug in the PHP SDKs. Additional concerns are PII and sensitive data. The best way to achieve this would be the use of attachments, but this is something we have to implement in an upcoming major version of the PHP SDK first.

For now, the file path is hopefully enough to debug issues.

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
cleptric: The filename, if accessible, would be a good description.
You could also construct the data array with prefixed keys, like `file.path` and such.

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * @template TResult
+     * @param array<string, mixed> $data
+     * @param callable(): TResult $originalCall
+     * @return TResult
+     */
+    protected function withTracing(string $method, array $data, callable $originalCall)
+    {
+        $context = new SpanContext();
+        $context->setOp("storage.{$method}");
```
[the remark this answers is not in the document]
spawnia: Done.

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
spawnia: Given that some operations work with directories or two paths, I think I am going to stick to just data to avoid redundant information.

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
spawnia: Thanks for the feedback, I documented the reasoning and implemented all methods accordingly.

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
spawnia: I made this throw.

config/sentry.php
```diff
@@ -54,6 +54,11 @@
         // Capture views as spans
         'views' => true,
 
+        // Capture storage access as spans
+        // Filesystems have to opt-in by changing their configuration like this:
+        // ['driver' => 'local', ...] -> ['driver' => 'sentry', 'original_driver' => 'local', ...]
```
spawnia, line 59: Instead of requiring manual configuration, we could do something like this in `StorageIntegration::setup()`:

```php
foreach (config('filesystem.disks') as $disk => $config) {
    config(["filesystem.disks.{$disk}.original_driver" => $config['driver']);
    config(["filesystem.disks.{$disk}.driver" => 'sentry');
}
```

What do you think?

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
cleptric: In general, we try to avoid the SDK breaking customer applications. While I agree that some issues might be hard to debug, especially in PHP, we must always degrade gracefully. In this case, just no-op.

spawnia: In this case, the application would fail without the SDK being present too.

src/Sentry/Laravel/Features/StorageIntegration.php
```diff
@@ -0,0 +1,35 @@
...
+    public function isApplicable(): bool
+    {
+        return $this->isTracingFeatureEnabled('storage');
+    }
+
+    public function setup(): void
+    {
+        $this->container()->afterResolving(FilesystemManager::class, static function (FilesystemManager $filesystemManager): void {
+            $filesystemManager->extend('sentry', function (Application $application, array $config) use ($filesystemManager): Filesystem {
+                $config['driver'] = $config['original_driver'];
```
stayallive, line 21: As a sanity check, maybe we should add a check and throw if the `original_driver` is for some reason `sentry` to prevent creating hard to debug infinite loops?

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /**
+     * @param list<mixed> $args
+     * @param array<string, mixed> $data
+     */
+    protected function withTracing(string $method, array $args, array $data)
+    {
+        $context = new SpanContext();
+        $context->setOp("file.{$method}"); // See https://develop.sentry.dev/sdk/performance/span-operations/#web-server
```
stayallive, line 36: It's actually under mobile, let's keep it generic.

```suggestion
        $context->setOp("file.{$method}"); // See https://develop.sentry.dev/sdk/performance/span-operations/
```

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
```diff
@@ -0,0 +1,183 @@
...
+use function Sentry\trace;
+
+/**
+ * Decorates the underlying filesystem by wrapping all calls to it with tracing.
+ *
+ * Parameters such as paths, directories or options are attached to the span as data,
+ * parameters that contain file contents are omitted due to potential problems with
+ * payload size or sensitive data.
+ */
+class TracingFilesystem implements CloudFilesystemContract
```
stayallive, line 19: I am wondering about this, we might want to have a seperate `TracingCloudFilesystem` and a `TracingFilesystem` calls. People might test for the implementing interface to see how a file system can be used.

stayallive (changes requested):

config/sentry.php
stayallive: This feel very hacky... like very 😅 not sure at this moment.

src/Sentry/Laravel/EventHandler.php
```diff
@@ -200,7 +200,7 @@ public function __call(string $method, array $arguments)
         }
 
         try {
-            call_user_func_array([$this, $handlerMethod], $arguments);
+            ($this->handlerMethod)(...$arguments);
```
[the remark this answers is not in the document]
spawnia: It is faster because it avoids a method call, see https://github.com/symfony/symfony/pull/29309.

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
spawnia: It is both under Web Server and Mobile, so I think we can be specific here.

stayallive: Woops, yes. I searched for `file.` 🤦 

Nothing was said!

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
spawnia: Laravel always implements `Cloud` and never just its parent, so I think we should be fine to do the same:

image

image

config/sentry.php
spawnia: I just tried it in my application and it seems to cause no issues so far, see [`4f322e0` (#726)](https://github.com/getsentry/sentry-laravel/pull/726/commits/4f322e080c14561c62baa893057bd5b378a24680).

Advantages to this approach:
- it just works™
- no manual configuration required
- no errors as mentioned in https://github.com/getsentry/sentry-laravel/pull/726#discussion_r1252182612 are possible

Disadvantages:
- feels scary

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
spawnia: Those packages seem fundamentally flawed in their assumption that this distinction is meaningful though, given that Laravel defaults to wrapping even local filesystems in `FilesystemAdapter` - which does implement `Cloud`.

`FilesystemAdapter` tries really hard to return a value for `url()`, even for local filesystems:

image

I think that issues with this would be very unlikely, but I am working on a solution to separate the implemenations.

src/Sentry/Laravel/Features/StorageIntegration.php
spawnia: I don't think we need to when we automate the driver swap with https://github.com/getsentry/sentry-laravel/pull/726#discussion_r1252201982.

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
spawnia: Done with [`f40bbbf` (#726)](https://github.com/getsentry/sentry-laravel/pull/726/commits/f40bbbf4bc92ad41c4794a9b7a4d797d23ddb10a).

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
spawnia: I actually found usability without description pretty poor, so I just added the JSON encoded data. Does not look to bad: 
image

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
cleptric: It's fine if an exception is thrown by the framework, we should just refrain from throwing one ourselves.

spawnia: I see, so we just try the call without checking? I guess it could still work due to duck typing, so we can let it slide.

spawnia:
I am running this version in our test environment since yesterday and have not had issues so far. The span looks good to me:

image

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
```diff
@@ -0,0 +1,164 @@
...
+        $context->setDescription(json_encode($data, JSON_PRETTY_PRINT));
+        $context->setData($data);
+
+        return trace(function () use ($method, $args) {
+            return $this->filesystem->{$method}(...$args);
+        }, $context);
+    }
+
+    /** @see \Illuminate\Filesystem\FilesystemAdapter::assertExists() */
+    public function assertExists($path, $content = null)
```
stayallive, line 43: Why would we want to add these if we already have the `__call` these are test only methods right don't see much point in tracing them?

stayallive (commented):
I do not like the way the description looks as a JSON string and I also don't think it adds much value doing it like that, so I've opted to add a dedicated description which is almost always the path that is being worked on:

image

(copy and move show to "from" path and when multiple paths are passed (for delete) I show the count of them)

Ofcourse the data hasn't changed, I did also add the driver name and disk name to the data since that seems like useful info if you use multiple filesystems.

I thing we should still remove the testing methods on our tracing filesystem since those are not part of the interface and don't add much since I don't think anyone enables Sentry in their test suite (I hope), agree?

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
spawnia: > these are test only methods right

I think that is their intended purpose, but that does not mean that everybody is going to use them that way. Given they are present in `Illuminate\Support\Facades\Storage`, I could see an uninformed developer using them in live code.

stayallive: Hmmmmmmm. Okay sure, they won't hurt I guess.

src/Sentry/Laravel/Features/StorageIntegration.php
stayallive: I disagree since stupid things can still happen :)

I've reworked it so users can't do uninteded things and we are also a little more cautious.

config/sentry.php
stayallive: Yep, I hear ya. Let's keep it like this!

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
```diff
@@ -0,0 +1,183 @@
...
+            $data = ['path' => $paths];
+            $description = $paths;
+        }
+
+        return $this->withTracing(__FUNCTION__, func_get_args(), $description, $data);
+    }
+
+    public function copy($from, $to)
+    {
+        return $this->withTracing(__FUNCTION__, func_get_args(), $from, compact('from', 'to'));
```
spawnia, line 131:
```suggestion
        return $this->withTracing(__FUNCTION__, func_get_args(), "From {$from} to {$to}", compact('from', 'to'));
```

Same with `move`, perhaps adding quotes around the paths

stayallive: On the one hand, yes, on the other it makes the description very long potentially... 

@cleptric WDYT?

cleptric: Right now, this looks like this for me.

Screenshot 2023-07-25 at 20 00 20

stayallive:
@spawnia I've made quite some changes, would appreciate a last look from you in case I missed something!

cleptric (approved):
Great job @spawnia & @stayallive 🚀

src/Sentry/Laravel/Features/StorageIntegration.php
```diff
@@ -0,0 +1,83 @@
...
+class StorageIntegration extends Feature
+{
+    private const FEATURE_KEY = 'storage';
+
+    private const STORAGE_DRIVER_NAME = 'sentry';
+
+    public function isApplicable(): bool
+    {
+        return $this->isBreadcrumbFeatureEnabled(self::FEATURE_KEY)
+               || $this->isTracingFeatureEnabled(self::FEATURE_KEY);
```
spawnia, line 22:
```suggestion
        return $this->isBreadcrumbFeatureEnabled(self::FEATURE_KEY)
            || $this->isTracingFeatureEnabled(self::FEATURE_KEY);
```

Forgive my nitpick-ness, I just think it is beatiful how `return` and a standard indentation of 4 spaces + `||` have the same length.

src/Sentry/Laravel/Features/StorageIntegration.php
```diff
@@ -0,0 +1,83 @@
...
+
+            config([
+                "filesystems.disks.{$disk}.driver" => self::STORAGE_DRIVER_NAME,
+                "filesystems.disks.{$disk}.sentry_disk_name" => $disk,
+                "filesystems.disks.{$disk}.sentry_original_driver" => $config['driver'],
+            ]);
+        }
+
+        $recordSpans = $this->isTracingFeatureEnabled(self::FEATURE_KEY);
+        $recordBreadcrumbs = $this->isBreadcrumbFeatureEnabled(self::FEATURE_KEY);
```
spawnia, line 42: I think fetching this information can be done at a later point in the callback of `extend`. This avoids performing unnecessary work eagerly. Of course this relies on the callbacks only executing once, which I believe to be the case. I tried this by adding two `Storage` calls to a route and putting an `echo` in either place, both run only once.

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
```diff
@@ -0,0 +1,210 @@
...
+    /**
+     * @param list<mixed> $args
+     * @param array<string, mixed> $data
+     */
+    protected function withTracing(string $method, array $args, string $description, array $data)
+    {
+        $op = "file.{$method}"; // See https://develop.sentry.dev/sdk/performance/span-operations/#web-server
+        $data = array_merge($data, $this->defaultData);
+
+        if ($this->recordBreadcrumbs) {
```
spawnia, line 50: Are breadcrumbs also considered tracing?

src/Sentry/Laravel/Util/Filesize.php
```diff
@@ -0,0 +1,32 @@
...
+     * @param int $dec
+     *
+     * @return string
+     */
+    public static function toHuman(int $bytes, int $dec = 2): string
+    {
+        $size   = ['B', 'kB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+        $factor = floor((strlen($bytes) - 1) / 3);
+
+        if ($factor == 0) {
```
spawnia, line 25:
```suggestion
        $factor = (int) floor((strlen($bytes) - 1) / 3);

        if ($factor === 0) {
```

src/Sentry/Laravel/Util/Filesize.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+class Filesize
+{
+    /**
+     * Convert bytes to human readable format.
+     *
+     * Credit: https://stackoverflow.com/a/23888858/1580028
+     *
+     * @param int $bytes
+     * @param int $dec
```
spawnia, line 16: What is this?

src/Sentry/Laravel/Util/Filesize.php
```diff
@@ -0,0 +1,32 @@
...
+    {
+        $size   = ['B', 'kB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+        $factor = floor((strlen($bytes) - 1) / 3);
+
+        if ($factor == 0) {
+            $dec = 0;
+        }
+
+        return sprintf("%.{$dec}f %s", $bytes / (1024 ** $factor), $size[$factor]);
+
```
spawnia, line 30:
```suggestion
```

spawnia (commented):
I like the addition of breadcrumbs and the improved descriptions.

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
spawnia: Perhaps the class and method should be renamed to something like `SentryFilesystem` and `withSentry` then.

stayallive (approved):
Thanks @spawnia, changes made!

spawnia (commented):
Nice! I have this branch deployed to our testing environment without issues so far, trying out the updated version now.

src/Sentry/Laravel/Tracing/Storage/TracingFilesystem.php
stayallive: You might be right yeah

stayallive:
You made a good point with the class naming, so I've updated that too now. Should be just moving stuff around, nothing functional :)

LukeTowers:
@spawnia @cleptric this is breaking Winter CMS installs using S3 storage drivers with the following error:

```
Winter\Storm\Filesystem\Filesystem::isLocalDisk(): Argument #1 ($disk) must be of type Illuminate\Filesystem\FilesystemAdapter,
Sentry\Laravel\Features\Storage\SentryCloudFilesystem given, called in
/var/task/vendor/laravel/framework/src/Illuminate/Support/Facades/Facade.php on line 338
```

This is because `isLocalDisk()` expects an `Illuminate\Filesystem\FilesystemAdapter` (see https://github.com/wintercms/storm/blob/develop/src/Filesystem/Filesystem.php#L164) which is getting automatically replaced by `Sentry\Laravel\Features\Storage\SentryCloudFilesystem` which doesn't extend Laravel's base adapter class.

This issue is caused simply by running `composer update`, no configuration changes were required to trigger this.

stayallive:
@LukeTowers, see #745.

LukeTowers:
Thanks @stayallive. For additional background as to why @wintercms type hints the `FilesystemAdapter` instead of the contract is because we're specifically making use of a method that is on that class and not on the contract (at least AFAIK), the `getAdapter()` method that returns the underlying Flysystem adapter instance.

stayallive:
Thanks @LukeTowers for the added information, it makes sense why'd someone want to do it. I guess this should eventually be added to the contracts or something if it's something people would commonly need to do.

spawnia:
> Is there a technical reason why the contract approach was chosen instead of inheritance?

One reason was to maintain separation between `\Illuminate\Contracts\Filesystem\Filesystem` and `\Illuminate\Contracts\Filesystem\Cloud`. However, I don't see why it could not be done, `Illuminate\Filesystem\FilesystemAdapter` is not `final`. It feels a bit dirty to me, but if *it just works™* - why not?

> doesn't this feature impact performance

Of course it does, as does all instrumentation Sentry does. However, I think Storage operations are likely to be several orders of magnitude slower than the few extra functions that are called to instrument them.

meyer59:
Hey 
Since this change we receive a lot of errors errors when acsessing to the Storage `Driver [sentry] is not supported.` 
   We diddn't change anything in our config files.  Can you check what's going on pls ?

PR #732, opened 2023-07-17
Less sensitive data by default

stayallive:
This PR changes the following:

- Do not show SQL query bindings in breadcrumbs by default
- Filter session cookie and remember cookies from the request by default

### DIFF
--- a/src/Sentry/Laravel/Http/LaravelRequestFetcher.php
+++ b/src/Sentry/Laravel/Http/LaravelRequestFetcher.php
@@ -3,6 +3,8 @@
 namespace Sentry\Laravel\Http;
 
 use Illuminate\Container\Container;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Str;
 use Psr\Http\Message\ServerRequestInterface;
 use Sentry\Integration\RequestFetcher;
 use Sentry\Integration\RequestFetcherInterface;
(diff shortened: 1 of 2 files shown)


src/Sentry/Laravel/Tracing/Middleware.php
```diff
@@ -142,7 +142,10 @@ public function setBootedTimestamp(?float $timestamp = null): void
 
     private function startTransaction(Request $request, HubInterface $sentry): void
     {
-        $requestStartTime = $request->server('REQUEST_TIME_FLOAT', microtime(true));
+        $requestStartTime = $request->server(
+            'REQUEST_TIME_FLOAT',
+            defined('LARAVEL_START') ? LARAVEL_START : microtime(true)
+        );
```
cleptric, line 148: Shall we do this in #734?

src/Sentry/Laravel/Http/LaravelRequestFetcher.php
```diff
@@ -26,9 +28,27 @@ public function fetchRequest(): ?ServerRequestInterface
...
         }
 
-        return (new RequestFetcher)->fetchRequest();
+        if ($request === null) {
+            return null;
+        }
+
+        $cookies = new Collection($request->getCookieParams());
+
+        $forbiddenCookies = [config('session.cookie'), 'remember_*'];
```
cleptric, line 42: Would slap a small comment here so our future selves remember why we did this 😃

cleptric (approved):

PR #736, opened 2023-07-25
Livewire 3 & Volt

stayallive:
This is tracking Livewire 3 support.

Known issues or observed problems:

- [x] Livewire request not detected (`x-livewire` header present but now empty)
- [ ] `component.booted` no longer fired or incorrectly listened to
- [ ] Check if changes needed to properly annotate Volt components

Let us know if you see anything else awry when using Livewire 3!

### DIFF
--- a/src/Sentry/Laravel/Features/LivewirePackageIntegration.php
+++ b/src/Sentry/Laravel/Features/LivewirePackageIntegration.php
@@ -137,7 +137,7 @@ class LivewirePackageIntegration extends Feature
                 return false;
             }
 
-            return $request->header('x-livewire') === 'true';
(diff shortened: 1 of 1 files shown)


stayallive:
Yes, absolutely, making sure we are ready but should not be released before 3.0 release tag 👍

stayallive:
Livewire 3 has made significant changes to their events and structure so our current setup is not working and will need a rewrite for version 3. There is no upgrade documentation provided by the Livewire team so we will have to do a reverse engineering/debug session to make the current code work with the new Livewire code (and keep Livewire 2 compat too).

stayallive:
Closing this in favor of #798.

PR #738, opened 2023-07-25
Laravel Folio support

stayallive:
For the PR to make sense we need this PR to be merged first: laravel/folio#34.

I don't think we need "folio" spans since Folio uses a normal route and blade view rendering it works out great with our current spans. We just need to adjust the transaction name to look like a normal route (first in image below) instead of a wildcard route (second in image below).

image

### Todo's

- [x] laravel/folio#34
- [x] Test with multiple segment routes
- [x] Test with model bound routes
- [x] Duplicated `route` and `route.folio` breadcrumbs (should not have original `route` breadcrumb)
- [x] Add tests


stayallive:
Agreed! Just have the PR up so we can try it out but let's wait for an official release before releasing this in the wild!

src/Sentry/Laravel/Features/FolioPackageIntegration.php
```diff
@@ -0,0 +1,59 @@
...
+
+        $transaction->setName($routeName);
+        $transaction->getMetadata()->setSource(TransactionSource::route());
+    }
+
+    private function extractRouteForMatchedView(MatchedView $matchedView, MountPath $mountPath): string
+    {
+        $path = Str::beforeLast('/' . ltrim($mountPath->baseUri . $matchedView->relativePath(), '/'), '.blade.php');
+
+        return str_replace(['[', ']'], ['{', '}'], $path);
```
cleptric, line 57: We could use `Str::replace` as Folio only supports Laravel 10 and 11.

cleptric (approved):
Nice one! 🪨

PR #740, opened 2023-07-26
Add PHPStan

cleptric:
Let's kick off things with level 0. We can start iterating on this.
The baseline ignores all optional packages we hook into in the SDK.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,7 +5,6 @@ on:
   push:
     branches:
       - master
-      - develop
       - release/**
 
 jobs:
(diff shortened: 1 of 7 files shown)


szepeviktor:
PHPStan is for really clean OOP.
Laravel is full of "Gimme a Closure I love that".

https://github.com/getsentry/sentry-laravel/blob/a21c479038dcaf9e295feb01e090dc4ac722a33e/src/Sentry/Laravel/Features/Storage/Integration.php#L61-L66

The comment says, it a hack.
The best way to resolve this is to redesign that solution!

szepeviktor:
> ignores all optional packages we hook into in the SDK.

The final solution for those packages are to install them as dev-deps just before static analysis and do analyze them. It takes a lot of time to develop that.

stayallive (approved):
💪

Issue #741, opened 2023-07-27
Request without Host header fails


georgeboot:
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

3.6.1

### Steps to reproduce

1. Create a laravel app with the sentry SDK and ensure you have a SDN set.
2. Sent a request WITHOUT a host header
3. You will get a InvalidArgumentException, thrown by the Psr7 factory.

### Expected result

It should ignore these things, as Sentry is not the primary thing we run.

Lost of load balancers do health requests without any headers etc. That's exactly how we got to the issue.

I've submitted https://github.com/Nyholm/psr7/issues/249 to Nyholm/psr7 but I am not sure if they will fix it, since it's probably a non-standard thing.

### Actual result

```
InvalidArgumentException: Unable to parse URI: "http://:8000/health?source=lb"

vendor/nyholm/psr7/src/Uri.php in Nyholm\Psr7\Uri::__construct at line 55
```

cleptric:
I couldn't reproduce this with a Laraverl 10 app using version 3.6.1 of the SDK.

```
curl -v -H 'Host:' http://127.0.0.1
```

This does not result in an exception.

Could you also show use the full stack trace, please? And any other things that could help us reproduce the issue.

georgeboot:
That is an EMPTY host header. The issue happens when there is NO host header.

cleptric:
Thanks for sharing the stack trace! Happy when people use our issue sharing feature 😄

Sorry if I might get this wrong, but to my knowledge, `-H` 'Host:' does remove the entire header.

```
*   Trying 127.0.0.1:80...
* Connected to 127.0.0.1 (127.0.0.1) port 80 (#0)
> GET / HTTP/1.1
> User-Agent: curl/7.88.1
> Accept: */*
```

The header seems to be not present anymore if I `var_dump` the `Symfony\Component\HttpFoundation\Request` as well.

```
  ["headers"]=>
  object(Symfony\Component\HttpFoundation\HeaderBag)#49 (2) {
    ["headers":protected]=>
    array(2) {
      ["user-agent"]=>
      array(1) {
        [0]=>
        string(11) "curl/7.88.1"
      }
      ["accept"]=>
      array(1) {
        [0]=>
        string(3) "*/*"
      }
    }
    ["cacheControl":protected]=>
    array(0) {
    }
  }
```

As you are using Octane, I have to set up a small test app.
I'll get back to you once I find out more!

[...]

PR #743, opened 2023-07-27
Handle exceptions raised while resolving the request

cleptric:
Let's not crash when we can't resolve the request.

Fix for #741

### DIFF
--- a/src/Sentry/Laravel/Http/SetRequestMiddleware.php
+++ b/src/Sentry/Laravel/Http/SetRequestMiddleware.php
@@ -35,8 +35,6 @@ class SetRequestMiddleware
     {
         try {
             return $container->make(ServerRequestInterface::class);
-        } catch (BindingResolutionException $e) {
-            // This happens if Laravel doesn't have the correct classes available to construct the PSR-7 object
         } catch (Throwable $e) {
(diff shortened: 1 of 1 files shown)


src/Sentry/Laravel/Http/SetRequestMiddleware.php
```diff
@@ -36,6 +37,8 @@ private function resolvePsrRequest(Container $container): ?ServerRequestInterfac
             return $container->make(ServerRequestInterface::class);
         } catch (BindingResolutionException $e) {
             // This happens if Laravel doesn't have the correct classes available to construct the PSR-7 object
```
cleptric, line 39: We could also solely rely on the catch block below.

stayallive (approved):

stayallive:
So I vaguely seem to remember there was a case where we swalled a `SuspiciousOperationException` (or similar) which is why we are specificaly listening for the `BindingResolutionException` (1d2b7a7b05402d39211ca6db97184da1aa988bd5) however I cannot seem to find the actual reason anywhere or what the problem is and I also can't seem to reproduce at all. This should be a safe change but wanted to document it here just in case it turned out not to be so we know where to look.

Issue #745, opened 2023-07-31
Class::disk(): Return value must be of type Illuminate\Filesystem\FilesystemAdapter, Sentry\Laravel\Features\Storage\SentryCloudFilesystem returned


Gummibeer:
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

3.7.0

### Steps to reproduce

Have a method type-hinting for `Illuminate\Filesystem\FilesystemAdapter` and return `Storage:disk('s3')` or any other cloud disk.

### Expected result

Type-hint before v3.7.0 still works as expected and a breaking change of the package i sonly introduced during a new major-release.

### Actual result

TypeError is thrown with message:
> Return value must be of type `Illuminate\Filesystem\FilesystemAdapter`, `Sentry\Laravel\Features\Storage\SentryCloudFilesystem` returned

atypic-garrett:
Came across the same issue, had to update composer to use 3.6.* :(

Gummibeer:
There isn't anything helpful in the stacktrace - as the problem is the new sentry storage breadcrumbs and so on. And the rebind of the storages already in place.
Because of that the type of the resolved driver changes and this destroys all existing type-hints. To resolve that you will have to go back to v3.6 or do the work and type-hint the underlying interface instead of the precise class.

cleptric:
In general, please assume that someone might not have all the necessary context to reproduce an issue immediately. In this case, a stack trace or your list of installed dependencies would have helped me verify that someone needs to install `league/flysystem-aws-s3-v3` to reproduce the issue.

I'm seeing the problem now, and we'll get this fixed momentarily 🙂

Gummibeer:
It's not related to S3 specifically.  It's a problem with every single fileystem and related to your `Sentry/Laravel/Features/Storage/Integration.php:40` logic which rebinds every single fileystem with a new precise class. Because of that every single place with a precise class type-hint will now fail.

stayallive:
@Gummibeer I understand this is not expected and we are looking for a solution other than just simply disabling the feature. 

But why are you type hinting the implementing class and not the interface (`Illuminate\Contracts\Filesystem\(Cloud)Filesystem`)?

stayallive:
[...]

PR #746, opened 2023-07-31
Disable storage integration by default

cleptric:
Potential fix for #745 

~~This will still cause issues if someone types the concrete class, though.~~

```php
$storage = fn(): Illuminate\Filesystem\AwsS3V3Adapter => Storage::disk('s3');
```

Edit: no longer a concern, disabled means disabled and we do nothing. Still has (same) impact when enabled manually, but this will be documented instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,12 @@ The Sentry SDK team is happy to announce the immediate availability of Sentry La
 
 ### Bug Fixes
 
-- Performance traces and breadcrumbs for filesystem access are now disabled by default [(#746)](https://github.com/getsentry/sentry-laravel/pull/746)
+- Performance traces and breadcrumbs for filesystem access are now turned off by default [(#746)](https://github.com/getsentry/sentry-laravel/pull/746)
 
   To enable the feature, you'll need to make some changes in your `config/filesystems.php` file for each disk where you want to enable the tracing filesystem driver.
   
   ```php
-  // For example, if you want to trace the `local` disk you update the disk config from this:
+  // For example, if you want to trace the `local` disk, you update the disk config from this:
   'local' => [
       'driver' => 'local',
       'root' => storage_path('app'),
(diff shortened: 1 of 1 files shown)


stayallive (approved):

spawnia (approved):

spawnia:
> This will still cause issues if someone types the concrete class, though.

You mean if enabled, right? If disabled, nothing should break.

stayallive:
@spawnia sorry, the PR description is a bit outdated, we first tried extending the Laravel Filesystem, but that won't work or solve all problems, the comment was for that solution. This will still break when you enable it (ofcourse not when disabled). We will document this instead since there is no way for us to transparantly replace the class without seeing the effects.

stayallive:
We've opted to move this feature to a "documentation only" feature, meaning you can still enable this but you need to change your filesystems config for it to do so. We will add documentation to [docs.sentry.io](https://docs.sentry.io/platforms/php/guides/laravel/performance/instrumentation/) on how to do this and also mention the possible problems you can run into when doing that.

src/Sentry/Laravel/Features/Storage/Integration.php
```diff
@@ -17,26 +17,12 @@ class Integration extends Feature
...
-            if ($currentDriver === self::STORAGE_DRIVER_NAME) {
-                continue;
-            }
-
-            config([
-                "filesystems.disks.{$disk}.driver" => self::STORAGE_DRIVER_NAME,
-                "filesystems.disks.{$disk}.sentry_disk_name" => $disk,
-                "filesystems.disks.{$disk}.sentry_original_driver" => $config['driver'],
-            ]);
-        }
```
spawnia, line 38: I would love to see this preserved in a `public static function configureDisksWithSentryDriver(): void` in this class. The documentation could mention that it can be used to simplify the configuration change.

cleptric: We can follow up on this later. For now, we want to get a fix out ASAP.

siarheipashkevich:
I've the same issue also. When are you going to fix this?

cleptric:
@siarheipashkevich https://github.com/getsentry/sentry-laravel/releases/tag/3.7.1

Issue #739, opened 2023-07-25
Shorten view paths in spans


stayallive:
### Problem Statement

When playing around with Laravel Folio I noticed the view spans have the full path the the views:

![image](https://github.com/getsentry/sentry-laravel/assets/1090754/387a264c-3937-425b-8952-ee3bf5ddf943)

This is a little on the long and verbose side.

### Solution Brainstorm

We should remove `base_path()` from those view paths turning:

```
/Users/alex/dev/sentry-laravel-test/resources/views/pages/users/[id].blade.php
```

In just:

```
/resources/views/pages/users/[id].blade.php
```

(since views can also be in the vendor folder and other places stripping away `resources/views` too seems a bit too much, so that's why I'm suggesting just the `base_path()`)

Issue #513, opened 2021-09-28
Errors not getting grouped when they are the same from inside blade files


olivernybroe:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
sentry.php.laravel 2.8.0

### Steps to Reproduce
Whenever an error happens from inside a blade file, they are added as a new issue, if the error was in a different version. (we clear blade cache on each release)
![image](https://user-images.githubusercontent.com/5870441/135047502-cec21b82-cca1-4354-8463-10e8c0e8d943.png)

When looking at the stacktrace, the reason it doesn't group them might be because of `param0`, which contains the hashed id of the cached blade file. This means when we clear the cache, that id will be different, and sentry will detect it as a different issue.
![image](https://user-images.githubusercontent.com/5870441/135048075-ab3ded63-dc40-43ef-9378-1a46d5c77ba8.png)

Another plausible reason could be that sentry reports that the stacktrace cannot be found for issues, as long as the issue happens inside a blade file. (warning is not shown in non blade files)

![image](https://user-images.githubusercontent.com/5870441/135047330-43fd1189-5a58-4599-a8bf-1518785c113e.png)

olivernybroe:
Looking at this a bit closer, I think this might be the fault of ignition, as ignition wraps all the blade exceptions into a `Facade\Ignition\Exceptions\ViewException`. 

So this might be solvable by removing ignition from production 🤔

stayallive:
Removing Ignition from production seems like a great plan anyway unless you have a specific reason for having it there to prevent having multiple error handles, moving it to `require-dev` should be enough if you are just using it for the pretty debug screens. 

The trace not being found should only happen if you are using performance tracing and the transaction is not or not correctly set or the trace was never sent (are you using sampling) and that might be a separate issue all together.

But removing Ignition will not solve it, we need to strip or normalize the view paths from the exception message for grouping to function correctly which we are not doing at the moment. The frame paths are correct but I actually suspect that is because of Ignition :)

I'll leave this open as a feature request, we need to both cleanup the exception message and the stack frame paths to show the actual blade file paths instead of the compiled paths.

---

I do have a "solution" in the meantime that fixes the exception messages but it ain't pretty and I did not had the time to package this up nicely for the SDK yet:

```php
<?php

namespace App\Exceptions;

use Sentry\Event;
use Sentry\ExceptionDataBag;

class Sentry
{
[...]

PR #747, opened 2023-08-01
Replace compiled view paths in profiles

stayallive:
This works with: getsentry/sentry-php#1574, and is a PoC to show what that PR could be used for.

It replaces compiled view paths with their non-compiled paths (the ones you would actually look at in your editor).

Todo:

- [ ] getsentry/sentry-php#1574
- [ ] Check and make sure we should replace `filename` / `function` and not more / other info
- [ ] When zero-downtime deployments are in play the compiled path can be outside the `base_path`, we might need to add `realpath(storage_path())` as a prefix path to fix this. This requires some more testing / investigating
- [ ] Possibly memoize the results to prevent multiple I/O operations for the same path? (uses a little memory)

### DIFF
--- a/src/Sentry/Laravel/Profiling/FrameProcessor.php
+++ b/src/Sentry/Laravel/Profiling/FrameProcessor.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Sentry\Laravel\Profiling;
+
+use Illuminate\Support\Str;
+use Sentry\Laravel\Tracing\BacktraceHelper;
+
+class FrameProcessor
+{
+    /**
+     * @param array $frame
+     *
+     * @return array
+     */
+    public function __invoke(array $frame): array
+    {
+        // Check if we are dealing with a frame for a cached view path
+        if (Str::startsWith($frame['filename'], '/storage/framework/views/')) {
+            $originalViewPath = $this->backtraceHelper()->getOriginalViewPathForCompiledViewPath($frame['abs_path']);
+
+            if ($originalViewPath !== null) {
+                // For views both the filename and function is the view path so we set them both to the original view path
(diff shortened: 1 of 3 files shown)


src/Sentry/Laravel/Profiling/FrameProcessor.php
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * @param array $frame
+     *
+     * @return array
+     */
+    public function __invoke(array $frame): array
+    {
+        // Check if we are dealing with a frame for a cached view path
+        if (Str::startsWith($frame['filename'], '/storage/framework/views/')) {
+            $originalViewPath = $this->backtraceHelper()->getOriginalViewPathForCompiledViewPath($frame['abs_path']);
```
cleptric, line 19: Could we benchmark the perf overhead?

cleptric:
This is becoming more relevant now, as we'll ship a new performance issue that alerts users about regressed function durations. Having the path to the actual view file would make these issues more actionable.

src/Sentry/Laravel/Profiling/FrameProcessor.php
stayallive: I think we could but not sure how to quantify, since this does a few I/O operations to read the view file. In this WIP I didn't implement memoizing which could reduce the amount of operations needed when a view is seen multiple times. I could run this code a few iterations on my machine and see what the result is, but it's always an increase since before we didn't do anything.

stayallive:
If we get this to a nice state we can also re-use the code for #513 and #739.

src/Sentry/Laravel/Profiling/FrameProcessor.php
cleptric: A rough ballpark would suffice.

PR #748, opened 2023-08-01
Move to PSR-4

szepeviktor:
Move from PSR-0 autoloading to PSR-4.

@cleptric The middle commit contains no content change.


stayallive:
Hey @szepeviktor, forgive my ignorance, but what is the benefit of this change?

The biggest downside of this is that `git blame` is going be a lot more difficult and although PSR-0 is deprecated and PSR-4 is now recommended I don't see anything about it no longer being supported anywhere so if this I feel this requires a little more justification.

szepeviktor:
Hello @stayallive 👋🏻 

I live very closely to software I am given, and I just couldn't live with PSR-0.
That is all.

PR #768, opened 2023-08-28
Add Laravel about command registration

SebastiaanKloos:
This PR adds the important configuration options to the Laravel about command.

https://laravel.com/docs/10.x/packages#about-artisan-command


stayallive (approved):
I like it, thanks for the PR! It currently looks like this:

image

PR #780, opened 2023-09-27
Fix deprecation message for str_starts_with

simbig:
str_starts_with(): passing null to parameter #1 ($haystack) of type string is deprecated in /var/www/vendor/laravel/framework/src/illuminate/support/str.php on line 1298

### DIFF
--- a/src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
+++ b/src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
@@ -80,7 +80,7 @@ class LighthouseIntegration implements IntegrationInterface
     private function processEvent(Event $event, Options $options): void
     {
         // Detect if we are processing a GraphQL request, if not skip processing the event
-        if (!Str::startsWith($event->getTransaction() ?? '', 'lighthouse?')) {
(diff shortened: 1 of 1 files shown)


src/Sentry/Laravel/Tracing/Integrations/LighthouseIntegration.php
```diff
@@ -80,7 +80,7 @@ public function setupOnce(): void
     private function processEvent(Event $event, Options $options): void
     {
         // Detect if we are processing a GraphQL request, if not skip processing the event
-        if (!Str::startsWith($event->getTransaction(), 'lighthouse?')) {
+        if (!Str::startsWith($event->getTransaction() ?? '', 'lighthouse?')) {
```
stayallive, line 83: How about we rewrite this as:

```suggestion
        if ($event->getTransaction() === null || !Str::startsWith($event->getTransaction(), 'lighthouse?')) {
```

That reads better to me and possibly saves some string operations that are not needed anyway.

stayallive (changes requested):
Thanks for making the PR, have 1 little suggestion!

cleptric (approved):

cleptric:
This was just released in [v3.8.1](https://github.com/getsentry/sentry-laravel/releases/tag/3.8.1)

PR #783, opened 2023-10-11
Fix `DateTimeZone` not properly converted to a string

### LINKED ISSUE #782: [Cron Monitoring] DateTimeZone Argument Error
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

3.8.1

### Steps to reproduce

We recently updated from Sentry version 3.3.3 to 3.8.1, and when we did our Cron monitor started throwing errors. We aren't doing anything fancy, but here's our command:
```
$schedule->command(LogScheduledTasksRunning::class)->everyMinute()->sentryMonitor();
```

### Expected result

Using the old version of the SDK had no issue running the `sentryMonitor()` command without issue.

We've found that passing `sentryMonitor(updateMonitorConfig: false)` fixes the issue.

### Actual result

This error is thrown:
```
TypeError 

Sentry\MonitorConfig::__construct(): Argument #4 ($timezone) must be of type ?string, DateTimeZone given, called in /var/www/html/api/vendor/sentry/sentry-laravel/src/Sentry/Laravel/Features/ConsoleIntegration.php on line 97

at vendor/sentry/sentry/src/MonitorConfig.php:29
   25▕      * @var string|null The timezone
   26▕      */
   27▕     private $timezone;
   28▕ 
➜  29▕     public function __construct(
   30▕         MonitorSchedule $schedule,
   31▕         ?int $checkinMargin = null,
   32▕         ?int $maxRuntime = null,
   33▕         ?string $timezone = null

   +25 vendor frames 
26  artisan:35
   Illuminate\Foundation\Console\Kernel::handle()
```

stayallive:
Fixes #782.

### DIFF
--- a/src/Sentry/Laravel/Features/ConsoleIntegration.php
+++ b/src/Sentry/Laravel/Features/ConsoleIntegration.php
@@ -2,6 +2,7 @@
 
 namespace Sentry\Laravel\Features;
 
+use DateTimeZone;
 use Illuminate\Console\Application as ConsoleApplication;
 use Illuminate\Console\Scheduling\Event as SchedulingEvent;
 use Illuminate\Contracts\Cache\Factory as Cache;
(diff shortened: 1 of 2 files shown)


cleptric (approved):
Thanks 😊

PR #785, opened 2023-10-16
Run CI on PHP 8.3

cleptric:


.github/workflows/ci.yaml
```diff
@@ -104,6 +104,13 @@ jobs:
           - php: "8.2"
             packages: { laravel: ^7.0,  testbench: 5.1.*, phpunit: 8.4.* }
 
+          - php: "8.3"
+            packages: { laravel: ^6.0,  testbench: 4.7.*, phpunit: 8.4.* }
+          - php: "8.3"
+            packages: { laravel: ^7.0,  testbench: 5.1.*, phpunit: 8.4.* }
+          - php: "8.3"
+            packages: { laravel: ^8.0,  testbench: ^6.0, phpunit: 9.3.* }
```
cleptric, line 112: Unsure about this. [phpspec/prophecy](https://github.com/phpspec/prophecy/issues/607) currently doesn't work with PHP 8.3.

stayallive: I'm perfectly fine not testing these ancient versions with `8.3` since the framework itself (on that version) also doesn't support that version. Also highly doubt that people running those versions still are also on `8.3`. If `prophecy` is updated after release of `8.3` and we can run the tests good, but otherwise not a big deal IMHO.

stayallive (approved):

PR #796, opened 2023-11-13
Prepare 4.0

cleptric:


cleptric:
We can also keep the older changelog entries if you want.

stayallive (approved):
🪨

stayallive:
Hah. Just noise. They are all available in the history if needed and also in the 3.x branch 👍

PR #797, opened 2023-11-13
Refactor HTTP client integration feature

stayallive:
This refactors the HTTP client integration to make use of the Feature structure and cleanup the code overal and increase test coverage. Some other goals:

- [x] Add `baggage` and `sentry-trace` headers to requests (using laravel/framework#47525)


stayallive:
Woops... I need to combine this with our current tracing setup to prevent duplicate spans, back to draft!

src/Sentry/Laravel/Features/HttpClientIntegration.php
```diff
@@ -0,0 +1,207 @@
...
+    {
+        if ($this->isTracingFeatureEnabled(self::FEATURE_KEY)) {
+            $events->listen(RequestSending::class, [$this, 'handleRequestSendingHandlerForTracing']);
+            $events->listen(ResponseReceived::class, [$this, 'handleResponseReceivedHandlerForTracing']);
+            $events->listen(ConnectionFailed::class, [$this, 'handleConnectionFailedHandlerForTracing']);
+
+            // The `globalRequestMiddleware` functionality was introduced in Laravel 10.14
+            if (method_exists($factory, 'globalRequestMiddleware')) {
+                $factory->globalRequestMiddleware([$this, 'attachTracingHeadersToRequest']);
+            }
```
cleptric, line 44: We have to move this out of this scope, as we always want to apply the headers.

stayallive: I figured it would be nice to be able to disable it, but I guess you already can using `trace_propagation_targets` of course so there is no need to gate it behind the tracing config 👍

cleptric (approved):

Issue #787, opened 2023-10-20
Livewire 3 requests are not identified as Livewire in transansaction


hmazter:
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

3.8.2

### Steps to reproduce

1. Use Laravel with Livewire 3 (regular Livewire Components, not Volt)
2. Make livewire component update request, using the default `POST /livewire/update` endpoint
3. See the transaction in Sentry as /livewire/update

### Expected result

I expect to see the component name in the transaction to differentiate between different component updates

### Actual result

I see the default `/livewire/update` as transaction name 
![Screenshot 2023-10-20 at 08 06 14](https://github.com/getsentry/sentry-laravel/assets/805377/e6884d28-5b0b-4b4f-bda8-4b1463601b0e)

I guess this is being addressed in #736

stayallive:
Yes, correct, unfortunately Livewire team has made some drastic changes and no guide on how to migrate our current setup to their new structure. So we didn't complete the work yet to get our integration working for Livewire 3. If you have insights I'd be glad to accept some help on the PR otherwise we will eventually get to it!

hmazter:
Thanks for the answer!
I'm pretty new to Livewire so I don't have any insight (yet) that could help 😄

cleptric:
Did some research today

We can still listen to some events via `LivewireManager::listen`

- `component.booted` -> ???
- `component.boot` -> ???
- `component.dehydrate` -> `dehydrate`
- `component.mount` -> `mount`

There are zero docs, zero upgrade guides nor any meaningful PRs to look at, so this is very tedious.

To check for a livewire request, they now perform `request()->hasHeader('X-Livewire')`.

cleptric:
Support for Livewire 3 was added in [v4.0.0](https://github.com/getsentry/sentry-laravel/releases/tag/4.0.0).

hoangtrunga1k55:
> Support for Livewire 3 was added in [v4.0.0](https://github.com/getsentry/sentry-laravel/releases/tag/4.0.0).

Is support for livewire2?

stayallive:
Yes, both Livewire 2 and 3 are supported.

hoangtrunga1k55:
@stayallive Can you help me?
[...]

PR #798, opened 2023-11-14
Added support for Livewire 3

SebastiaanKloos:
Fixes https://github.com/getsentry/sentry-laravel/issues/787#issuecomment-1795061390


composer.json
stayallive, line 1: We do this mostly for the types, so we can get autocomplete working and make phpstan a little happier without many exclusions.

src/Sentry/Laravel/Features/LivewirePackageIntegration.php
```diff
@@ -32,11 +31,56 @@ public function isApplicable(): bool
     }
 
     public function onBoot(LivewireManager $livewireManager): void
+    {
+        if (class_exists(EventBus::class)) {
```
stayallive, line 35: There is no version number, so we check for the `EventBus` class introduced in Livewire v3.

src/Sentry/Laravel/Features/LivewirePackageIntegration.php
```diff
@@ -125,6 +202,12 @@ private function updateTransactionName(string $componentName): void
         $transaction->getMetadata()->setSource(TransactionSource::custom());
 
         Integration::setTransaction($transactionName);
+
+        $hub->configureScope(function (Scope $scope) {
+            $livewireManager = $this->container()->make(LivewireManager::class);
+
+            $scope->setTag('livewire.original_url', $livewireManager->originalUrl());
```
cleptric, line 209: SDKs should not send any tags by default. Is this something a user needs to be able to search for?

SebastiaanKloos: @cleptric Livewire sends state updates to a global Livewire route (`/livewire/update`) and is not very helpful to find out the exact location of an issue. The original url helps to find out on what page the issue occurred. We already improved this issue by renaming the transaction to include the Livewire component. Livewire components can be used on many pages so adding this tag makes it much easier to track down the origin of the issue.

cleptric: If the transaction name is already fixed, the `transaction` value of an error should reflect this as well. So a user will see issues grouped on a route basis afaik.

stayallive: We've decided to remove this for now and see in a follow up if we can add this parameter somewhere and make it searchable still. But we want to ship this so we'll leave it out for now.

cleptric (approved):

stayallive (approved):
💪

cleptric:
Thanks @SebastiaanKloos and @stayallive!

PR #799, opened 2023-11-14
Update `sentry/sdk` dependency

tenshiAMD:
By updating `sentry/sdk` to `v4.0` which uses [`sentry/sentry` now removed `php-http/message-factory` dependency](https://github.com/getsentry/sentry-php/blob/df4c078a349f588fc7c704d66f3c46c26487d9e2/CHANGELOG.md#misc). Therefore, we can get rid of this dependency issue when installing this package.
```
Package `php-http/message-factory` is abandoned, you should avoid using it. Use psr/http-factory instead.
```


cleptric:
Version 4.0 of the PHP SDK contains breaking changes, so it's not done by just bumping it.
We're working on a new major version of the Laravel SDK which will be released soon.

Issue #801, opened 2023-11-17
In Tracing, is there consideration to add `sql.bindings` to `queryExecuted`?


summerKK:
### Problem Statement

Is there consideration to add a `db.bindings` to record SQL bindings, making it convenient for SQL analysis? 
It could be set as a configurable option for enabling.
```php
    protected function queryExecutedHandler(DatabaseEvents\QueryExecuted $query): void
    {
        if (!$this->traceSqlQueries) {
            return;
        }

        $parentSpan = SentrySdk::getCurrentHub()->getSpan();

        // If there is no tracing span active there is no need to handle the event
        if ($parentSpan === null) {
            return;
        }

        $context = new SpanContext();
        $context->setOp('db.sql.query');
        $context->setDescription($query->sql);
        $context->setData([
            'db.name' => $query->connection->getDatabaseName(),
            'db.system' => $query->connection->getDriverName(),
            'server.address' => $query->connection->getConfig('host'),
            'server.port' => $query->connection->getConfig('port'),
        ]);
        $context->setStartTimestamp(microtime(true) - $query->time / 1000);
        $context->setEndTimestamp($context->getStartTimestamp() + $query->time / 1000);

        if ($this->traceSqlQueryOrigins) {
            $queryOrigin = $this->resolveQueryOriginFromBacktrace();

            if ($queryOrigin !== null) {
                $context->setData(array_merge($context->getData(), [
                    'db.sql.origin' => $queryOrigin
                ]));
            }
        }

        $parentSpan->startChild($context);
    }
```

### Solution Brainstorm

```php
if ($this->traceSqlBindings) {
    $context->setData(array_merge($context->getData(), [
        'db.sql.bindings' => $query->bindings,
    ]));
}
```

stayallive:
It already is:

https://github.com/getsentry/sentry-laravel/blob/a5120ccd3fa1641badc92695bc3dd18ca242fa16/config/sentry.php#L46-L47

Is that option not doing what you are asking here?

summerKK:
I mean in the tracing span, there is no binding value now.

cleptric:
We may add this in the future. PRs are welcome!

summerKK:
Thanks for accepting my `MR`

PR #804, opened 2023-11-21
Adding bindings to SQL spans in tracing

summerKK:
I submitted a PR for #801.  😉

### DIFF
--- a/src/Sentry/Laravel/Tracing/EventHandler.php
+++ b/src/Sentry/Laravel/Tracing/EventHandler.php
@@ -42,6 +42,13 @@ class EventHandler
     private $traceSqlQueries;
 
     /**
+     * Indicates if we should add query bindings to query spans.
+     *
+     * @var bool
+     */
+    private $recordSqlBindings;
+
+    /**
      * Indicates if we should we add SQL query origin data to query spans.
      *
      * @var bool
@@ -83,6 +90,7 @@ class EventHandler
     {
         $this->traceSqlQueries = ($config['sql_queries'] ?? true) === true;
         $this->traceSqlQueryOrigins = ($config['sql_origin'] ?? true) === true;
+        $this->recordSqlBindings = ($config['sql_bindings'] ?? true) === true;
 
         $this->traceQueueJobs = ($config['queue_jobs'] ?? false) === true;
         $this->traceQueueJobsAsTransactions = ($config['queue_job_transactions'] ?? false) === true;
@@ -174,6 +182,12 @@ class EventHandler
                     'db.sql.origin' => $queryOrigin
                 ]));
             }
+        }
+
+        if ($this->recordSqlBindings) {
+            $context->setData(array_merge($context->getData(), [
+                'db.sql.bindings' => $query->bindings
+            ]));
         }
 
         $parentSpan->startChild($context);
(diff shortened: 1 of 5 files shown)


src/Sentry/Laravel/Tracing/EventHandler.php
```diff
@@ -176,6 +184,12 @@ protected function queryExecutedHandler(DatabaseEvents\QueryExecuted $query): vo
             }
         }
 
+        if ($this->recordSqlBindings) {
+            $context->setData(array_merge($context->getData(), [
+                'db.sql.bindings' => $query->bindings
+            ]));
+        }
```
cleptric, line 191: Can you move this above `traceSqlQueryOrigins`, so we have the same order?

src/Sentry/Laravel/Tracing/EventHandler.php
```diff
@@ -83,6 +90,7 @@ public function __construct(array $config)
     {
         $this->traceSqlQueries = ($config['sql_queries'] ?? true) === true;
         $this->traceSqlQueryOrigins = ($config['sql_origin'] ?? true) === true;
+        $this->recordSqlBindings = ($config['sql_bindings'] ?? true) === true;
```
cleptric, line 93: Can you move this below `$this->traceSqlQueries`?

src/Sentry/Laravel/Tracing/EventHandler.php
```diff
@@ -176,6 +184,12 @@ protected function queryExecutedHandler(DatabaseEvents\QueryExecuted $query): vo
             }
         }
 
+        if ($this->recordSqlBindings) {
+            $context->setData(array_merge($context->getData(), [
+                'db.sql.bindings' => $query->bindings
```
cleptric, line 189: @AbhiPrasad any better ideas for the naming here?

cleptric (commented):
Thank you, left some super nitpicky comments 😄

src/Sentry/Laravel/Tracing/EventHandler.php
stayallive: I've gone ahead and made those changes 👍

src/Sentry/Laravel/Tracing/EventHandler.php
stayallive: I've gone ahead and made those changes 👍

stayallive (approved):
I've cleaned up the tests a bit since we already had some stuff in place for testing queries. But otherwise looking good, thanks!

Let's see if we need to adjust the `db.sql.bindings` key but otherwise 🚢

src/Sentry/Laravel/Tracing/EventHandler.php
AbhiPrasad: Not solved in opentelemetry unfortunately - https://github.com/open-telemetry/opentelemetry-specification/pull/3092

The JS otel sdk has `db.postgresql.values` https://github.com/epsagon/opentelemetry-js-contrib/blob/1f6299d9d48ef2f7b206170827d3858e947474db/plugins/node/opentelemetry-instrumentation-pg/src/enums/AttributeNames.ts#L18

I would be open to `db.sql.values`, but that does make this more general, maybe the constrained definition of `db.sql.bindings` is good here.

Issue #805, opened 2023-11-22
Cron monitors not working


gregory-claeyssens:
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

4.0.0

### Steps to reproduce

Following the documentation to the letter by adding `->sentryMonitor()`
Example:

```
$schedule->command('cache:prune-stale-tags')->hourly()->sentryMonitor();
```

### Expected result

It simply work

### Actual result

An error is logged in Sentry
`Method Illuminate\Console\Scheduling\Event::sentryMonitor does not exist.`

cleptric:
We need a bit more info, please, like the Laravel version. Is anything custom in terms of SDK setup? Anything else you can think of?

rzv-me:
I got the same error. 

We are using Laravel Vapor.
The problem, from what I understand from the code, is that Laravel Vapor instantiates the console kernel
`$app->make(ConsoleKernelContract::class)->call('config:cache');` in `vendor/laravel/vapor-core/stubs/fpmRuntime.php`, but the `isApplicable()` function on `ConsoleIntegration`, returns true only when running in console.

My solution right now is to add 
```
    protected function addSentryMonitorMacro()
    {
        if(!SchedulingEvent::hasMacro('sentryMonitor')) {
            SchedulingEvent::macro('sentryMonitor', function (
                ?string $monitorSlug = null,
                ?int $checkInMargin = null,
                ?int $maxRuntime = null,
                bool $updateMonitorConfig = true
            ) {
                return $this;
            });
        }
    }
```
to the Console Kernel and to run it at the beggining of the `schedule()` method: `$this->addSentryMonitorMacro();`

cleptric:
@stayallive can you take a look please?

rzv-me:
I submitted a PR for this, the better solution I think is to do it in the `register()` method of `ConsoleIntegration`

cleptric:
Fixed in [4.1.1](https://github.com/getsentry/sentry-laravel/releases/tag/4.1.1)

PR #812, opened 2023-11-29
Fix missing `sentryMonitor` macro when command is called outside the CLI environment

rzv-me:
This is a PR to fix the problem described by the #805 issue.

The problem is that whenever an artisan command is called when not running in console, the schedule method on the ConsoleKernel is executed. 

This happens on Laravel Vapor, for example, where the config is cached on the fly with `$app->make(ConsoleKernelContract::class)->call('config:cache');` in `vendor/laravel/vapor-core/stubs/fpmRuntime.php`

this can be replicated easily, by doing `Artisan::call('inspire');` in a controller.

### DIFF
--- a/src/Sentry/Laravel/Features/ConsoleIntegration.php
+++ b/src/Sentry/Laravel/Features/ConsoleIntegration.php
@@ -28,6 +28,11 @@ class ConsoleIntegration extends Feature
      */
     private $cache;
 
+    public function register(): void
+    {
+        $this->onBootInactive();
+    }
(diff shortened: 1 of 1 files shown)


stayallive (approved):
This is a awkward situation but I guess it's not the worst. Been thinking if there was another way we could better fix this. I guess we would need a `bootWhenNotApplicable` or something otherwise. But since the inactive version only adds a simple closure to an array this is fine for now. Thanks!

PR #819, opened 2023-12-12
Add `cache.hit` to Redis span data

stayallive:
See: https://develop.sentry.dev/sdk/performance/span-data-conventions/#web-server

This is not ideal, but I think this is the best we can currently do since we don't have any cache spans other than these unless we decorate the cache store (which is a future project to tackle).


stayallive:
This should be done for all cache spans not just Redis, closing this in favor of a better solution.

PR #820, opened 2023-12-17
Queue tracing bugfixes

stayallive:
Fixes the following:

- Moves the queue tests to a more logical place, and adds tests for the queue job tracing feature flag
- Fixes bug where feature tests are incorrectly memoized in the feature class (this goes for all feature classes, not just the queue although it was only noticeable there since it had multiple options of the same category)
- ~~Fixes `trace` context being lost for `queue.process` transactions because the context was overwritten by the default context because there was no active span in the scope which is required for the correct context being added to the transaction event~~ (fixed in PHP SDK instead)

Especially the last point is important, this fixes data to get lost for queue job transactions, the before: 

image

and after the change:

image

this also fixes the missing operation name in the performance dashboard (top one old, bottom one new) and missing data on the transaction itself (the queue job name and attempts and such):

image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
             "@phpstan",
             "@tests"
         ],
-        "tests": "vendor/bin/phpunit --verbose",
+        "tests": "vendor/bin/phpunit",
         "cs-check": "vendor/bin/php-cs-fixer fix --verbose --diff --dry-run",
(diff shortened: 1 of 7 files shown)


stayallive:
getsentry/sentry-php#1668 fixes part of this PR, back to draft to validate and update.

cleptric (approved):

PR #822, opened 2023-12-20
Laravel Lumen performance traces not collected because route never matched

### LINKED ISSUE #821: Lumen does not have the `Illuminate\Routing\Events\RouteMatched event`.
### How do you use Sentry?

Self-hosted / on-premises

### SDK version

4.1

### Steps to reproduce

Lumen uses `nikic/fast-route` instead of `illuminate/routing`. This event will not be triggered, causing `tracing` information to be unable to be reported. `$this->didRouteMatch` is always `false`

vendor/sentry/sentry-laravel/src/Sentry/Laravel/Tracing/Middleware.php
```php
  public function terminate(Request $request, $response): void
  {
      // If there is no transaction or the HubInterface is not bound in the container there is nothing for us to do
      if ($this->transaction === null || !$this->app->bound(HubInterface::class)) {
          return;
      }

      // We stop here if a route has not been matched unless we are configured to trace missing routes
      if (!$this->didRouteMatch && config('sentry.tracing.missing_routes', false) === false) {
          return;
      }
  }
```

If it's `lumen`, can we just determine `didRouteMatch` based on `$response->getStatusCode() === 404`

### Expected result

Report normally

### Actual result

Unable to report

summerKK:
Fixes #821.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -162,10 +162,10 @@ parameters:
 
 		-
 			message: "#^Class Laravel\\\\Lumen\\\\Application not found\\.$#"
-			count: 2
-			path: src/Sentry/Laravel/Tracing/ServiceProvider.php
+			count: 1
+			path: src/Sentry/Laravel/Tracing/Middleware.php
 
 		-
-		    message: "#^Class Laravel\\\\Lumen\\\\Application not found\\.$#"
-		    count: 1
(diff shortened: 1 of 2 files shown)


stayallive (approved):
I changed to logic to what we had before we started using the event to detect if a route was matched and moved it to a helper for clearer code and made sure missing routes are still traced also for Lumen. Thanks for taking the time to report and submit a PR!

cleptric (approved):

PR #825, opened 2023-12-21
Refactor model violation reporters

stayallive:
This refactors the way model violation reports to cleanup the code and remove duplication where possible.

Also adds memoization (enabled by default) to prevent multiple reports in the same request.

Depends on #824.


cleptric (approved):
Do we have docs for this?

PR #833, opened 2024-01-17
Add span for publishing a job to the queue

stayallive:
With the (very) recent laravel/framework#49722 we can now add a span for jobs being queued 🔥 

afbeelding

### DIFF
--- a/src/Sentry/Laravel/Features/QueueIntegration.php
+++ b/src/Sentry/Laravel/Features/QueueIntegration.php
@@ -88,8 +88,9 @@ class QueueIntegration extends Feature
         $context = (new SpanContext)
             ->setOp('queue.publish')
             ->setData([
-                'job' => $jobName,
-                'connection' => $event->connectionName,
+                'messaging.system' => 'laravel',
+                'messaging.laravel.job' => $jobName,
+                'messaging.destination.connection' => $event->connectionName,
             ])
             ->setDescription($jobName);
 
--- a/src/Sentry/Laravel/Features/QueueIntegration.php
+++ b/src/Sentry/Laravel/Features/QueueIntegration.php
@@ -2,10 +2,13 @@
 
 namespace Sentry\Laravel\Features;
 
+use Closure;
 use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Queue\Events\JobExceptionOccurred;
 use Illuminate\Queue\Events\JobProcessed;
 use Illuminate\Queue\Events\JobProcessing;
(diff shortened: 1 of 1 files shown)


src/Sentry/Laravel/Features/QueueIntegration.php
```diff
@@ -62,6 +68,43 @@ public function onBoot(Dispatcher $events): void
...
+            $jobName = 'Closure';
+        } elseif (is_object($jobName)) {
+            $jobName = get_class($jobName);
+        }
+
+        $context = (new SpanContext)
+            ->setOp('queue.publish')
+            ->setData([
+                'job' => $jobName,
+                'connection' => $event->connectionName,
```
AbhiPrasad, line 92: Would be nice if we can fill in `messaging.system` from https://opentelemetry.io/docs/specs/semconv/attributes-registry/messaging/

Otherwise we probably want to use `messaging.destination.name` instead of `job` and maybe find the OTEL equivalent for `connectionName`? If we can't find anything for `connectionName` I'm fine to keep it.

stayallive: `messaging.system` doesn't have a well-known one for Laravel but I can set it to `laravel` which seems to be allowed.

The `job` is the name of the handler (usually a class) and not the name of the queue it's published on, does that still make sense to use `messaging.destination.name` for that? Eventually I'd like to add the actual queue name too but we need some changes on the framework side to surface that and we might want to reserve the destination name for that info.

Maybe the connection could be in `messaging.destination.connection` (not-standard). In Laravel you can have multiple queue connections configured like `sqs` or `redis` or a fake `sync` driver and the connection name indicates which queue connection it was pushed to.

To give an example of how a Laravel queue message might look like:

```
job: App\Jobs\SendSomeEmail
queue: mailer
payload: <JSON>
connectionName: redis
```

AbhiPrasad: > The job is the name of the handler (usually a class) and not the name of the queue it's published on, does that still make sense to use messaging.destination.name for that

Ahhh okay, then maybe we namespace to `laravel` to have it be less generic?

> Maybe the connection could be in messaging.destination.connection

Makes sense to me, let's do that change.

stayallive: Okay, I've made those changes, the op set to `queue.publish` is good?

AbhiPrasad (approved):
Conventions LGTM ✅

cleptric (approved):

Issue #836, opened 2024-01-23
[Cron Monitoring] Whether to consider adding a switch configuration option for Cron monitoring?


summerKK:
### Problem Statement

The scenario goes like this.
If I run in `dev environment`, `php artisan schedule:run`, `Cron` side will generate `CheckIn`, resulting in a lot of `CheckIn`. Can you consider adding a configuration option to avoid `dev environment` to enable this function?

### Solution Brainstorm

sentry.php

```php
[
    'crons' => [
        'monitor_enabled' => env('SENTRY_CRONS_MONITOR_ENABLED', false),
    ],
]
```

src/Sentry/Laravel/Features/ConsoleIntegration.php
```php
public function isApplicable(): bool
{
    return $this->container()->make(Application::class)->runningInConsole() && $this->isCronsFeatureEnabled('monitor_enabled');
}
```

summerKK:
I created a PR 😎
 #837

cleptric:
`before_send_check_in` was added to the PHP SDK in [v4.5.0](https://github.com/getsentry/sentry-php/releases/tag/4.5.0).

PR #837, opened 2024-01-23
Add a switch configuration option for Cron monitoring

summerKK:
for #836


summerKK:
> You can set the `DSN` to `null`. It has the same effect. I'm not in favour of adding yet another option.

Sometimes may need the `DSN` locally for debugging, and then after a while you forget to clear the `DSN` to execute the `schedule:run`, and that's a tragedy!😭

cleptric:
We'll add a new `before_send_check_in` option. This will allow you to conditional disable sending check-ins. https://github.com/getsentry/sentry-php/pull/1690

PR #842, opened 2024-01-29
Improve `sentry:test` command

stayallive:
- Show any internal non-debug/info messages always
- Show any log message when running with `-v`
- Add early check if cURL extension is enabled


src/Sentry/Laravel/Console/TestCommand.php
```diff
@@ -46,6 +46,12 @@ class TestCommand extends Command
      */
     public function handle(): int
     {
+        if (!\extension_loaded('curl')) {
+            $this->error('The cURL PHP extension must be enabled.');
```
cleptric, line 50:
```suggestion
            $this->error('You need to enable the PHP cURL extension (ext-curl).');
```

PR #846, opened 2024-02-01
First class callable helper for integrating Sentry with Laravel 11

stayallive:
Using: https://www.php.net/manual/en/functions.first_class_callable_syntax.php new in 8.1+ which we can recommend to use because Laravel 11 is PHP 8.1+ 🥳 

**Not tested yet!**

### DIFF
--- a/src/Sentry/Laravel/Integration.php
+++ b/src/Sentry/Laravel/Integration.php
@@ -59,7 +59,7 @@ class Integration implements IntegrationInterface
     public static function handles(Exceptions $exceptions): void
     {
         $exceptions->reportable(static function (Throwable $exception) {
-            self::captureUnhandledException($throwable);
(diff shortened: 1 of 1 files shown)


src/Sentry/Laravel/Integration.php
```diff
@@ -52,6 +53,16 @@ public function setupOnce(): void
         });
     }
 
+    /**
+     * Convienence method to register the exception handler with Laravel 11+.
+     */
+    public static function handles(Exceptions $exceptions): void
+    {
+        $exceptions->reportable(static function (Throwable $throwable) {
```
cleptric, line 61:
```suggestion
        $exceptions->reportable(static function (Throwable $exception) {
```

cleptric (approved):

PR #850, opened 2024-02-06
Make `queue.publish` spans the parent of `queue.process` spans

stayallive:
This makes way more sense in the performance trace and also gives us access to a little more data (the queue name) and also allows `queue.process` to include the serializing of the job payload.


cleptric (approved):
🚀

Issue #625, opened 2022-12-14
Email performance spans and/or breadcrumbs


stayallive:
We could listen to the `Illuminate\Mail\Events\MessageSending` and `Illuminate\Mail\Events\MessageSent` events to figure out when an mail was sent and possibly nest the view rendering needed to sent that e-mail neatly underneath a `mail` span.

We could potentially also add a breadcrumb indicating an email was sent by the framework.

ankurk91:
@stayallive 
Any plans for automated cron job monitoring?

https://docs.sentry.io/product/crons/getting-started/
https://laravel.com/docs/9.x/scheduling#events

cleptric:
@ankurk91 Scheduled task tracing was released in [`4.11.0`](https://github.com/getsentry/sentry-laravel/releases/tag/4.11.0).

PR #852, opened 2024-02-09
Implement notifications tracing and breadcrumbs

stayallive:
This implements (partially) #625.

Laravel also has a `Illuminate\Notifications\Events\NotificationFailed`, however this event is actually unused because when a notification fails to sent it just throws an exception from the underlying transport. So no need to handle this event.


stayallive:
Seems like I do not to investigate a bit why tests are reporting so many errors on Laravel 9 and below (don't think the code is wrong but something in the tests).

cleptric (approved):
Docs and :shipit:

stayallive:
https://github.com/getsentry/sentry-docs/pull/9674

PR #857, opened 2024-03-03
feat: add sentry tunnel controller

Muetze42:
Add optional Controller to tunnel blocked client requests.

Reference: https://github.com/getsentry/sentry-javascript/issues/2916

```php
// routes/api.php

Route::post('sentry-tunnel', \Sentry\Laravel\Http\Controllers\SentryTunnelController::class);
```

Vue Example:
```js
Sentry.init({
    app,
    dsn: import.meta.env.VITE_SENTRY_DSN_PUBLIC,
    tunnel: '/api/sentry-tunnel',
    trackComponents: true,
    logErrors: true
})
```

### DIFF
--- a/src/Sentry/Laravel/Http/Controllers/SentryTunnelController.php
+++ b/src/Sentry/Laravel/Http/Controllers/SentryTunnelController.php
@@ -1,0 +1,67 @@
+<?php
+
+namespace Sentry\Laravel\Http\Controllers;
+
+use GuzzleHttp\Client;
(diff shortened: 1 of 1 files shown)


cleptric:
If you'd like, you can contribute this snippet to the docs https://docs.sentry.io/platforms/javascript/troubleshooting/. Your example won't work with self-hosted Sentry installations, and I don't want to add this to the SDK at the moment.

PR #869, opened 2024-03-21
Add current Laravel context to event

### LINKED ISSUE #862: Add support for Laravel 11.0's Context
We should look at Laravel 11.0's new **context** feature.
Applying the Laravel **context** automatically to all events might be a good first step.
The same might apply to queue tracing, where **context** could allow us to implement trace propagation better.

Refs https://github.com/laravel/framework/pull/49730

stayallive:
Laravel introduces [context](https://laravel.com/docs/11.x/context) which is similar to how our scopes work, this introduces a new integration that adds (if present) the current context to the event/transaction data so they become visible in Sentry.

This can look something like this in the Sentry UI:

afbeelding

Fixes #862.

### DIFF
--- a/src/Sentry/Laravel/Integration/LaravelContextIntegration.php
+++ b/src/Sentry/Laravel/Integration/LaravelContextIntegration.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Sentry\Laravel\Integration;
+
+use Illuminate\Support\Facades\Context;
+use Sentry\Event;
+use Sentry\EventHint;
+use Sentry\EventType;
+use Sentry\Integration\IntegrationInterface;
+use Sentry\SentrySdk;
+use Sentry\State\Scope;
+
+class LaravelContextIntegration implements IntegrationInterface
+{
+    public function setupOnce(): void
+    {
+        // Context was introduced in Laravel 11 so we need to check if we can use it otherwise we skip the event processor
+        if (!class_exists(Context::class)) {
+            return;
+        }
+
+        Scope::addGlobalEventProcessor(static function (Event $event, ?EventHint $hint = null): Event {
+            $self = SentrySdk::getCurrentHub()->getIntegration(self::class);
+
+            if (!$self instanceof self) {
+                return $event;
+            }
+
+            if (!in_array($event->getType(), [EventType::event(), EventType::transaction()], true)) {
+                return $event;
+            }
+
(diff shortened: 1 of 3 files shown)


src/Sentry/Laravel/Integration/LaravelContextIntegration.php
```diff
@@ -0,0 +1,40 @@
...
+                return $event;
+            }
+
+            if (!in_array($event->getType(), [EventType::event(), EventType::transaction()], true)) {
+                return $event;
+            }
+
+            if (!Context::isEmpty()) {
+                $event->setContext('laravel', Context::all());
+            }
```
timacdonald, line 35: `Context::isEmpty` will consider both the "public" and "hidden" information.

I would instead recommend checking that there is any non-hidden data to attach.

```suggestion
            if (Context::all()) {
                $event->setContext('laravel', Context::all());
            }
```

test/Sentry/Integration/LaravelContextIntegrationTest.php
```diff
@@ -0,0 +1,72 @@
...
+        Context::flush();
+        Context::add('foo', 'bar');
+        Context::addHidden('hidden', 'value');
+    }
+
+    private function assertContextIsCaptured(array $context): void
+    {
+        $this->assertArrayHasKey('laravel', $context);
+        $this->assertArrayHasKey('foo', $context['laravel']);
+        $this->assertArrayNotHasKey('hidden', $context['laravel']);
```
timacdonald, line 69: Love that you have an assertion ensuring you aren't collecting hidden data ❤️

src/Sentry/Laravel/Integration/LaravelContextIntegration.php
stayallive: Turns out we don't even have to check at all since `setContext` already discards empty contexts being set so removed the check alltogether!

test/Sentry/Integration/LaravelContextIntegrationTest.php
stayallive: Yeah, better not capture those if we can help it 👍

src/Sentry/Laravel/Integration/LaravelContextIntegration.php
timacdonald: 🤩

PR #881, opened 2024-04-11
Limit cases where SQL origin is captured

stayallive:
Adds option to only capture origin for queries slower than a threshold (100 ms by default).

Also limit the frames to 20 when trying to detect an in-app frame that is considered the origin for an event.


src/Sentry/Laravel/Features/Concerns/ResolvesEventOrigin.php
```diff
@@ -16,7 +16,7 @@ protected function resolveEventOrigin(): ?string
     {
         $backtraceHelper = $this->makeBacktraceHelper();
 
-        $firstAppFrame = $backtraceHelper->findFirstInAppFrameForBacktrace(debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS));
+        $firstAppFrame = $backtraceHelper->findFirstInAppFrameForBacktrace(debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 20));
```
cleptric, line 19: A quick comment why `20` would be nice to have for future us 🙂

src/Sentry/Laravel/Tracing/EventHandler.php
```diff
@@ -194,13 +202,13 @@ protected function queryExecutedHandler(DatabaseEvents\QueryExecuted $query): vo
...
      *
-     * @return string|null
+     * @return array|null
      */
     private function resolveQueryOriginFromBacktrace(): ?array
     {
         $backtraceHelper = $this->makeBacktraceHelper();
 
-        $firstAppFrame = $backtraceHelper->findFirstInAppFrameForBacktrace(debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS));
+        $firstAppFrame = $backtraceHelper->findFirstInAppFrameForBacktrace(debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 20));
```
cleptric, line 211: Here as well.

cleptric (approved):
Docs and :shipit: 

Besides https://docs.sentry.io/platforms/php/guides/laravel/configuration/laravel-options/, there is also https://docs.sentry.io/product/performance/queries/. Can be one PR.

PR #892, opened 2024-05-03
Spotlight improvements

stayallive:
Two things happening:

- When Spotlight was enabled you also needed to set a DSN for the SDK to be enabled, now only enabling Spotlight is enough
- You couldn't enable Spotlight without adding something to your `config/sentry.php` now you can enable it from your `.env` using `SENTRY_SPOTLIGHT=true`
- Deprecated `couldHavePerformanceTracingEnabled` since it's no longer in use since #600

### DIFF
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -11,7 +11,7 @@ return [
     'dsn' => env('SENTRY_LARAVEL_DSN', env('SENTRY_DSN')),
 
     // @see https://spotlightjs.com/
-    'spotlight' => env('SENTRY_SPOTLIGHT', false),
+    // 'spotlight' => env('SENTRY_SPOTLIGHT', false),
 
     // @see: https://docs.sentry.io/platforms/php/guides/laravel/configuration/options/#logger
     // 'logger' => Sentry\Logger\DebugFileLogger::class, // By default this will log to `storage_path('logs/sentry.log')`
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -10,6 +10,9 @@ return [
     // @see https://docs.sentry.io/product/sentry-basics/dsn-explainer/
     'dsn' => env('SENTRY_LARAVEL_DSN', env('SENTRY_DSN')),
 
+    // @see https://spotlightjs.com/
+    'spotlight' => env('SENTRY_SPOTLIGHT', false),
+
     // @see: https://docs.sentry.io/platforms/php/guides/laravel/configuration/options/#logger
     // 'logger' => Sentry\Logger\DebugFileLogger::class, // By default this will log to `storage_path('logs/sentry.log')`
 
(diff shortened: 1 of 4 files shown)


config/sentry.php
```diff
@@ -10,6 +10,9 @@
     // @see https://docs.sentry.io/product/sentry-basics/dsn-explainer/
     'dsn' => env('SENTRY_LARAVEL_DSN', env('SENTRY_DSN')),
 
+    // @see https://spotlightjs.com/
+    'spotlight' => env('SENTRY_SPOTLIGHT', false),
+
```
cleptric, line 15: Let's treat this like the logger, as we don't want this to be enabled on prod.
So I would suggest no env var and commented out.

src/Sentry/Laravel/BaseServiceProvider.php
```diff
@@ -42,12 +42,14 @@ protected function getUserConfig(): array
      *
      * Because of `traces_sampler` being dynamic we can never be 100% confident but that is also not important.
      *
+     * @deprecated This method is deprecated and will be removed in the next major release.
```
cleptric, line 45:
```suggestion
     * @deprecated since version 4.6. To be removed in version 5.0.
```

src/Sentry/Laravel/BaseServiceProvider.php
```diff
@@ -22,7 +22,7 @@ protected function hasDsnSet(): bool
     {
         $config = $this->getUserConfig();
 
-        return !empty($config['dsn']);
+        return !empty($config['dsn']) || ($config['spotlight'] ?? false) === true;
     }
```
cleptric, line 26: What about we leave `hasDsnSet` as is and add `hasSpotlightEnabled` and centralize the checks in one additional method? Bit overkill, but cleaner. Up to you! 🙂

cleptric (approved):

config/sentry.php
stayallive: It's up to you, but not having an `.env` means you _have_ to publish the config files even though almost everything else is configurable without the config file published (it would be best if we wouldn't need users to publish the config file at all). Just add's an extra step to enable it locally, but no super hard feelings either way, so up to you.

cleptric: So far we told users during on-boarding to publish their config.
Let's revisit this in the next major, though there might be some surprises when we add new things to the SDK config and people get it automatically.

You can leave it as is then.

src/Sentry/Laravel/Tracing/ServiceProvider.php
```diff
@@ -27,7 +27,7 @@ class ServiceProvider extends BaseServiceProvider
     public function boot(): void
     {
         // If there is no DSN set we register nothing since it's impossible for us to send traces without a DSN set
```
cleptric, line 29: This needs some update, too

PR #901, opened 2024-05-30
fix: Avoid manipulating config when resolving disks

hansnn:
This commit updates the storage integration such that it does not manipulate the filesystems config when a disk is resolved.
Doing so fixes an issue when using scoped disks where the disk's prefix property is injected into the parent disk's config, making the parent disk unusable from that point forward.

The added test doesn't replicate the issue with scoped disks specifically because the `league/flysystem-path-prefixing` package is required to use scoped disks. The best I could do was to assert that the config isn't being manipulated when a disk is resolved, which seems sensible in and of itself.

Tagging you here, @spawnia, as you may have some insight into why the "hack" this commit removes was added in the first place, and if it's still needed (ref https://github.com/getsentry/sentry-laravel/pull/726).

### DIFF
--- a/src/Sentry/Laravel/Features/Storage/Integration.php
+++ b/src/Sentry/Laravel/Features/Storage/Integration.php
@@ -49,10 +49,14 @@ class Integration extends Feature
                     $diskResolver = (function (string $disk, array $config) {
                         // This is a "hack" to make sure that the original driver is resolved by the FilesystemManager
                         $oldConfig = config("filesystems.disks.{$disk}");
+
                         config(["filesystems.disks.{$disk}" => $config]);
+
                         /** @var FilesystemManager $this */
                         $resolved = $this->resolve($disk);
+
                         config(["filesystems.disks.{$disk}" => $oldConfig]);
+
                         return $resolved;
                     })->bindTo($filesystemManager, FilesystemManager::class);
 
--- a/src/Sentry/Laravel/Features/Storage/Integration.php
+++ b/src/Sentry/Laravel/Features/Storage/Integration.php
@@ -47,11 +47,8 @@ class Integration extends Feature
                     unset($config['sentry_original_driver']);
 
                     $diskResolver = (function (string $disk, array $config) {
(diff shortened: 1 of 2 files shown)


hansnn:
Seeing that the CI failed I now understand why the hack was added in the first place: The `$config` parameter is not available in `FilesystemManager::resolve` in Laravel < 8 (it was probably added in some version 8.X.X). I guess this fix can't be applied as-is, but I'll leave the PR open for now while investigating if there is a workaround to make it compatible with Laravel < 8 (unless you think it should be closed).

stayallive:
It would be nice if scoped disks will work correctly of course but we should also make sure that it works for all Laravel versions of course. So I'm happy to see if you can find a workable workaround to make this work for all supported versions.

hansnn:
The only solutions I can think of are to copy the `resolve` method from the Laravel library and use that directly instead of calling `$this->resolve()`, or to use reflection to check if the method accepts the `$config` parameter.

I'm not sure if there is any willingness to include either of those options. What do you think, @stayallive?

hansnn:
Pushed a commit that should work with all Laravel versions without using reflection or copying code from the Laravel repository. Kudos to @grevenx who came up with the solution.

src/Sentry/Laravel/Features/Storage/Integration.php
```diff
@@ -48,10 +48,12 @@ function (Application $application, array $config) use ($filesystemManager): Fil
 
                     $diskResolver = (function (string $disk, array $config) {
                         // This is a "hack" to make sure that the original driver is resolved by the FilesystemManager
```
spawnia, line 50: This was originally added by @stayallive in https://github.com/getsentry/sentry-laravel/pull/726/commits/81c5692b8c65436ca0fdf1a6de8cf2305267bf85.

spawnia (approved):
Looks like a sensible change to me.

stayallive (approved):
Thanks for taking the time to make the PR 👍

PR #902, opened 2024-06-03
Add support for the Queue Module

cleptric:
image

### DIFF
--- a/src/Sentry/Laravel/Features/QueueIntegration.php
+++ b/src/Sentry/Laravel/Features/QueueIntegration.php
@@ -36,6 +36,7 @@ class QueueIntegration extends Feature
 
     private const QUEUE_PAYLOAD_BAGGAGE_DATA = 'sentry_baggage_data';
     private const QUEUE_PAYLOAD_TRACE_PARENT_DATA = 'sentry_trace_parent_data';
+    private const QUEUE_PAYLOAD_PUBLISH_TIME = 'sentry_publish_time';
 
     public function isApplicable(): bool
     {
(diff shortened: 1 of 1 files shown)


src/Sentry/Laravel/Features/QueueIntegration.php
```diff
@@ -181,12 +180,22 @@ public function handleJobProcessingQueueEvent(JobProcessing $event): void
...
-            'connection' => $event->connectionName,
+            'messaging.system' => 'laravel',
+            
+            'messaging.destination.name' => $event->job->getQueue(),
+            'messaging.destination.connection' => $event->connectionName,
+            
+            'messaging.message.id' => (string) $event->job->getJobId(),
+            'messaging.message.envelope.size' => mb_strlen($event->job->getRawBody()),
+            'messaging.message.body.size' => mb_strlen(json_encode($event->job->payload()['data'])),
+            'messaging.message.retry.count' => $event->job->attempts(),
```
cleptric, line 197: We might want to decrement this by one, as this would be one by default.

stayallive (approved):

src/Sentry/Laravel/Features/QueueIntegration.php
```diff
@@ -181,12 +180,22 @@ public function handleJobProcessingQueueEvent(JobProcessing $event): void
...
-            'attempts' => $event->job->attempts(),
-            'connection' => $event->connectionName,
+            'messaging.system' => 'laravel',
+            
+            'messaging.destination.name' => $event->job->getQueue(),
+            'messaging.destination.connection' => $event->connectionName,
+            
+            'messaging.message.id' => (string) $event->job->getJobId(),
+            'messaging.message.envelope.size' => mb_strlen($event->job->getRawBody()),
+            'messaging.message.body.size' => mb_strlen(json_encode($event->job->payload()['data'])),
```
stayallive, line 196: You want to count bytes here, `mb_*` will count multibyte characters as a single byte instead. `strlen` can be used to count bytes ([see](https://www.php.net/manual/en/function.strlen.php#112411)).

```suggestion
            'messaging.message.envelope.size' => strlen($event->job->getRawBody()),
            'messaging.message.body.size' => strlen(json_encode($event->job->payload()['data'])),
```

stayallive (approved):

PR #904, opened 2024-06-04
Queue integration fixes

stayallive:
- Fixes: "ErrorException: Undefined array key "sentry_publish_time""
- Decode payload only once to save on CPU cycles
- Use job UUID from the payload (`$job->getJobId()` is publisher (like SQS) ID)
- Add job ID to publish span

### DIFF
--- a/src/Sentry/Laravel/Features/QueueIntegration.php
+++ b/src/Sentry/Laravel/Features/QueueIntegration.php
@@ -68,6 +68,7 @@ class QueueIntegration extends Feature
                         ->setOp(self::QUEUE_SPAN_OP_QUEUE_PUBLISH)
                         ->setData([
                             'messaging.system' => 'laravel',
+                            'messaging.message.id' => $payload['uuid'] ?? null,
                             'messaging.destination.name' => $queue,
                             'messaging.destination.connection' => $connection,
                         ])
--- a/src/Sentry/Laravel/Features/QueueIntegration.php
+++ b/src/Sentry/Laravel/Features/QueueIntegration.php
@@ -180,22 +180,19 @@ class QueueIntegration extends Feature
 
         $resolvedJobName = $event->job->resolveName();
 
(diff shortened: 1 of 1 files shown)


src/Sentry/Laravel/Features/QueueIntegration.php
```diff
@@ -68,6 +68,7 @@ public function onBoot(Dispatcher $events): void
                         ->setOp(self::QUEUE_SPAN_OP_QUEUE_PUBLISH)
                         ->setData([
                             'messaging.system' => 'laravel',
+                            'messaging.message.id' => $payload['uuid'] ?? null,
                             'messaging.destination.name' => $queue,
```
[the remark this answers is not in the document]
stayallive: Turns out... only the Redis driver does this. We could argue Laravel should fix this but let's also fix it ourselfs 😄

src/Sentry/Laravel/Features/QueueIntegration.php
```diff
@@ -180,22 +185,19 @@ public function handleJobProcessingQueueEvent(JobProcessing $event): void
...
         $job = [
             'messaging.system' => 'laravel',
-            
+
             'messaging.destination.name' => $event->job->getQueue(),
             'messaging.destination.connection' => $event->connectionName,
-            
-            'messaging.message.id' => (string) $event->job->getJobId(),
+
+            'messaging.message.id' => $jobPayload['uuid'] ?? $event->job->getJobId(),
```
cleptric, line 196: We need to keep the string cast for `$event->job->getJobId()`.

cleptric (approved):

src/Sentry/Laravel/Features/QueueIntegration.php
stayallive: It's documented as returning a string? When doesn't it do that?

PR #913, opened 2024-06-26
Supplemental elucidation for the documentation.

summerKK:
Greetings, in alignment with the upgrade documentation for Laravel, it is advised not to modify the directory structure when upgrading from `Laravel 10` to `Laravel 11`. Hence, I believe that most projects will retain their existing structure. Therefore, I have included additional clarifications to prevent any perplexity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,9 @@ Install the `sentry/sentry-laravel` package:
 composer require sentry/sentry-laravel
 ```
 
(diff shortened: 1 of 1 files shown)


cleptric:
Right at the top we mention:

> The installation steps below work on version 11.x of the Laravel framework. For older Laravel versions and Lumen see...

To keep our on-boarding concise across our various channels, I'm holding off on this proposal.

PR #932, opened 2024-07-24
Allow embedding SQL query bindings (parameters) in SQL query breadcrumbs and spans

spawnia:
I added a similar feature to Laravel in https://github.com/laravel/framework/pull/52192.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+### Features
+
+- Allow embedding SQL query bindings (parameters) in SQL query breadcrumbs and spans.
+
+  To enable this feature, update your `config/sentry.php` file or set the `SENTRY_TRACE_SQL_BINDINGS_ENABLED` and `SENTRY_BREADCRUMBS_SQL_BINDINGS_ENABLED` environment variables to `embed`.
+
+  ```php
+  'breadcrumbs' => [
+      'sql_bindings' => 'embed',
+  ],
+  'tracing' => [
+      'sql_bindings' => 'embed',
+  ],
+  ```
+
 ## 4.7.1
 
 The Sentry SDK team is happy to announce the immediate availability of Sentry Laravel SDK v4.7.1.
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -58,7 +58,7 @@ return [
         // Capture SQL queries as breadcrumbs
         'sql_queries' => env('SENTRY_BREADCRUMBS_SQL_QUERIES_ENABLED', true),
 
-        // Capture SQL query bindings (parameters) in SQL query breadcrumbs
+        // Capture bindings in SQL query breadcrumbs. `true` or `append` shows them separate, `embed` inlines them
         'sql_bindings' => env('SENTRY_BREADCRUMBS_SQL_BINDINGS_ENABLED', false),
 
         // Capture queue job information as breadcrumbs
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
 # Changelog
 
+### Features
+
+- Allow embedding SQL query bindings (parameters) in SQL query breadcrumbs and spans.
+
+  To enable this feature, update your `config/sentry.php` file or set the `SENTRY_TRACE_SQL_BINDINGS_ENABLED` and `SENTRY_BREADCRUMBS_SQL_BINDINGS_ENABLED` environment variables to `embed`.
```
spawnia, line 7: I considered the addition of an additional boolean configuration option for this, but found that this complicated the code and made it unclear which option had precedence over the existing options. However, I could see reverting back to making the options just a boolean and simply adopting the new behaviour?

config/sentry.php
```diff
@@ -58,7 +58,7 @@
         // Capture SQL queries as breadcrumbs
         'sql_queries' => env('SENTRY_BREADCRUMBS_SQL_QUERIES_ENABLED', true),
 
-        // Capture SQL query bindings (parameters) in SQL query breadcrumbs
+        // Capture bindings in SQL query breadcrumbs. `true` or `append` shows them separate, `embed` inlines them
```
spawnia, line 61: I shortened the prior part of the description to fit everything on a single short line still.

cleptric (changes requested):
This breaks our [Querries Insights Module](https://docs.sentry.io/product/insights/queries/), as it only works with parameterized queries.

We could debate adding the raw query to span data, but I currently don't see it that useful given the amount of changes required.

spawnia:
I think it would be massively useful to aid readability. I personally find it hard to decipher what is going on in the following example:

image

The inlined version is much more readable to me.

```sql
update `mll_app_exa_type_hdr_link` set `expt_valid_to` = 20240724140225, `expt_change_user` = 'bfranke' where `expt_id` = '62A2C0AB-D321-4C6F-9675-4EF2AC77BD34' and `expt_valid_to` = 99991231235555
```

Queries Insights works with spans right? I would be more interested in breadcrumbs anways, maybe we could just make the change there and omit the extra option for spans?

cleptric:
This is something we can fix in the product, we do not need to add more logic to the SDK. The data is already there. cc @gggritso

spawnia:
There are individual differences depending on the database or its configuration in how bindings are embedded, which is why the code that builds the raw SQL involves the `Connection` and `Grammar` classes. So I do not think that Sentry will be able to correctly build the SQL string from the data that is given in breadcrumbs. Would you consider a pull request that just offers the embedding for breadcrumbs? It should require fewer changes and provide the most value without breaking anything.

cleptric:
There are also concerns with server-side Pii stripping. Scrubbing individual data attributes is much easier than scrubbing a string, and I'm not sure we currently scrub breadcrumb descriptions. So a lot of reasons why I don't feel too happy about this change.

spawnia:
Are bindings such as the ones in the screenshot in https://github.com/getsentry/sentry-laravel/pull/932#issuecomment-2248061390 scrubbed? Seems difficult, given it is in the form of a list without keys that could indicate sensitive data.

spawnia:
Understood. Thanks for providing detailed feedback and discussing this pull request with me. I might try formulating a feature request over at https://github.com/getsentry/sentry.

cleptric:
We already looking into this, will hack something together next week. It’s a bit more involved but we want to have this feature 🙂

PR #933, opened 2024-07-24
Fix test class namespaces

spawnia:
Fixes the namespaces of test classes to conform with PSR-4.


stayallive:
Test suite is not liking this change.

spawnia:
> Test suite is not liking this change.

I rebased atop the latest `master`. It seems like only some versions are failing, I don't really get why. Do you have an idea what could be going on?

spawnia:
Found and fixed the reason for the failing tests.

stayallive (approved):
Cool, appreciate it, thanks!

PR #936, opened 2024-07-30
Move delayed transaction finisher state out of middleware

### LINKED ISSUE #935: 4.7.1 stops reporting performance units when using Laravel Octane
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

4.7.1

### Laravel version

10.48.17

### Steps to reproduce

1. Update from v4.7.0 to v4.7.1

### Expected result

For it to work as normal.

### Actual result

Sentry immediately stops reporting performance units after updating to 4.7.1. 

Image

Exceptions are still reported. This only affected my performance units.

Likely caused by changes in #921 

Using Octane 2.5.2 w/ Swoole. 

Reverting back to v4.7.0 immediately fixes the issue.

stayallive:
Because out middleware is a singleton resolved in the boot of the framework it never get's detroyed so any state remains forever, this caused the terminating callback to only be registered for one request and subsequent requests did not get a new terminating callback registered. To fix this I moved the registering of the callback to a class we keep as a singleton so we know if the container was reset and this ensure the callback is only registered once per container lifecycle and re-registers it as needed for Octane.

Fixes #935

This also works fine on normal Laravel applications. Lumen never used this method anyway since `app()->terminating()` isn't a thing there.

### DIFF
--- a/src/Sentry/Laravel/Tracing/TransactionFinisher.php
+++ b/src/Sentry/Laravel/Tracing/TransactionFinisher.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Sentry\Laravel\Tracing;
+
+/**
+ * @internal
+ */
+class TransactionFinisher
+{
+    public function __construct()
+    {
+        // We need to finish the transaction after the response has been sent to the client
+        // so we register a terminating callback to do so, this allows us to also capture
+        // spans that are created during the termination of the application like queue
+        // dispatched using dispatch(...)->afterResponse(). This middleware is called
+        // before the terminating callbacks so we are 99.9% sure to be the last one
+        // to run except if another terminating callback is registered after ours.
+        app()->terminating(function () {
+            app(Middleware::class)->finishTransaction();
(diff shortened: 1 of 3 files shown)


src/Sentry/Laravel/Tracing/TransactionFinisher.php
```diff
@@ -0,0 +1,22 @@
...
+class TransactionFinisher
+{
+    public function __construct()
+    {
+        // We need to finish the transaction after the response has been sent to the client
+        // so we register a terminating callback to do so, this allows us to also capture
+        // spans that are created during the termination of the application like queue
+        // dispatched using dispatch(...)->afterResponse(). This middleware is called
+        // before the terminating callbacks so we are 99.9% sure to be the last one
+        // to run except if another terminating callback is registered after ours.
```
stayallive, line 17: This comment needs an update, references "this middleware"

cleptric (approved):

PR #937, opened 2024-08-01
Fix config formatting

zepfietje:
This PR fixes formatting in the config file to follow Laravel's default Pint config.

Since Pint is installed in new Laravel projects by default, and the Sentry config is published after installation, running the code style fixer in CI would always lead to a "fix" commit. It's a minor thing, but still wanted to contribute this small "fix".


cleptric (approved):
Thanks!

PR #940, opened 2024-08-14
Set `http` breadcrumb level based on response code

cleptric:
image

I think it makes sense to set 4XX response codes to `warning` instead of `error`.

Part of https://github.com/getsentry/team-sdks/issues/100


stayallive (approved):
Makes sense to me!

PR #942, opened 2024-08-23
Allow cache store used by console scheduling integration to be overridden

jnbeaver:
An initial discussion about the implementation of this change can be found [here](https://discord.com/channels/621778831602221064/1268962575098449921).

- Allows the cache store used by `ConsoleSchedulingIntegration` to be overridden in a user-defined service provider
- Uses that cache store for storing state when a job is run in the background

### DIFF
--- a/src/Sentry/Laravel/Features/ConsoleSchedulingIntegration.php
+++ b/src/Sentry/Laravel/Features/ConsoleSchedulingIntegration.php
@@ -18,6 +18,9 @@ use Sentry\SentrySdk;
 
 class ConsoleSchedulingIntegration extends Feature
 {
+    /**
+     * @var string|null
+     */
     private $cacheStore = null;
 
     /**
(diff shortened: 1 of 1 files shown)


src/Sentry/Laravel/Features/ConsoleSchedulingIntegration.php
```diff
@@ -17,6 +18,8 @@
 
 class ConsoleSchedulingIntegration extends Feature
 {
+    private $cacheStore = null;
```
stayallive, line 21: My only suggestion, otherwise this looks good to me!

```suggestion
	/*
	 * @var string|null
	 */
    private $cacheStore = null;
```

stayallive (approved):

src/Sentry/Laravel/Features/ConsoleSchedulingIntegration.php
jnbeaver: Thanks, added!

cleptric (approved):

jnbeaver:
@stayallive @cleptric thanks for the reviews! Any idea when you're planning to cut the next release?

cleptric:
We'll get a new release out next week.

cleptric:
This is now available in [4.9.0](https://github.com/getsentry/sentry-laravel/releases/tag/4.9.0).

PR #954, opened 2024-11-23
Prevent PHP 8.4 deprecation warning

ejunker:
Prevent PHP 8.4 deprecation for default null value without nullable type declaration


cleptric:
Fixed in [4.10.1](https://github.com/getsentry/sentry-laravel/releases/tag/4.10.1).

PR #957, opened 2024-12-05
Compatibility with non default redis manager

Kriso1337:
change for compatibility with external redis manager (e.g. sentinel library in my case)

### DIFF
--- a/src/Sentry/Laravel/Features/CacheIntegration.php
+++ b/src/Sentry/Laravel/Features/CacheIntegration.php
@@ -6,6 +6,7 @@ use Illuminate\Cache\Events;
 use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Redis\Events as RedisEvents;
 use Illuminate\Contracts\Redis\Factory as RedisFactoryContract;
+use Illuminate\Redis\RedisManager;
 use Illuminate\Support\Str;
 use Sentry\Breadcrumb;
 use Sentry\Laravel\Features\Concerns\ResolvesEventOrigin;
@@ -60,7 +61,8 @@ class CacheIntegration extends Feature
         if ($this->isTracingFeatureEnabled('redis_commands', false)) {
             $events->listen(RedisEvents\CommandExecuted::class, [$this, 'handleRedisCommands']);
 
-            $this->container()->afterResolving(RedisFactoryContract::class, static function (RedisFactoryContract $redis): void {
+            $this->container()->afterResolving(RedisFactoryContract::class, static function ($redis): void {
+                /** @var RedisManager $redis */
                 $redis->enableEvents();
             });
         }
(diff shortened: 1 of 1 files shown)


src/Sentry/Laravel/Features/CacheIntegration.php
```diff
@@ -60,7 +60,7 @@ public function onBoot(Dispatcher $events): void
         if ($this->isTracingFeatureEnabled('redis_commands', false)) {
             $events->listen(RedisEvents\CommandExecuted::class, [$this, 'handleRedisCommands']);
 
-            $this->container()->afterResolving(RedisManager::class, static function (RedisManager $redis): void {
+            $this->container()->afterResolving(RedisFactoryContract::class, static function (RedisFactoryContract $redis): void {
                 $redis->enableEvents();
```
stayallive, line 64: The Factory does not contains `enableEvents` so we can't make this change as is: https://github.com/laravel/framework/blob/7db7ea950d3df663391718730f4490a4e189234a/src/Illuminate/Contracts/Redis/Factory.php.

stayallive (changes requested):
Is you sentinel `RedisManager` implementation open-source or from a package?

Kriso1337:
> Is you sentinel `RedisManager` implementation open-source or from a package?

from [package](https://packagist.org/packages/myqaa/laravel-redis-sentinel-drivers), it sounds like we can implement new interface here or maybe omit RedisManager type(and then add it to phpdoc)?

stayallive:
Thanks for linking the package, looks like this will work for that specific package.

> it sounds like we can omit RedisManager type(and then add it to phpdoc)?

I don't understand what you mean with this? Can you elaborate?

Kriso1337:
> Thanks for linking the package, looks like this will work for that specific package.
> 
> > it sounds like we can omit RedisManager type(and then add it to phpdoc)?
> 
> I don't understand what you mean with this? Can you elaborate?

sure, here it is, with this realization it pass any RedisManager, default and any with implemented contract, so we can use custom manager

or maybe I didn't understand you and you are worried about another problem?

stayallive:
So the main concern is that Sentry is installed in many applications and if we ask to container to give us an object of the `Illuminate\Contracts\Redis\Factory` type it can be any object. Not only the default Laravel one or the one from the Sentinel package. So it isn't guaranteed that `enableEvents` is there to call. In fact the contract guarantees it is not, but it might be.

Since there is no real good workaround this we are not making this change at this time. You can add the following snippet to your own `AppServiceProvider` which has the same effect and the Sentry SDK will be able to use the events to show the Redis operations as if it enabled the events itself:

```php
$this->app->afterResolving(\Monospice\LaravelRedisSentinel\RedisSentinelManager::class, static function (\Monospice\LaravelRedisSentinel\RedisSentinelManager $redis): void {
    $redis->enableEvents();
});
```

PR #977, opened 2025-02-03
Try harder to generate a slug for a cron monitor

stayallive:
Feedback from Laracon EU: We should have enough data to generate a slug for a scheduled job cron. So this PR tries a bit harder to generate something meaningful.

For job classes we reverse it to get the most interesting part of the name (the actual class name) in the front.

Also added a limit on our side to prevent jobs being not accepted (correctly) by sentry if the slug needs to be truncated. Current limit is 128.

### DIFF
--- a/src/Sentry/Laravel/Features/ConsoleSchedulingIntegration.php
+++ b/src/Sentry/Laravel/Features/ConsoleSchedulingIntegration.php
@@ -74,8 +74,8 @@ class ConsoleSchedulingIntegration extends Feature
             ?int $recoveryThreshold = null
         ) use ($startCheckIn, $finishCheckIn) {
             /** @var SchedulingEvent $this */
-            if ($monitorSlug === null && $this->command === null) {
-                throw new RuntimeException('The command string is null, please set a slug manually for this scheduled command using the `sentryMonitor(\'your-monitor-slug\')` macro.');
+            if ($monitorSlug === null && empty($this->command) && empty($this->description)) {
+                throw new RuntimeException('The command and description are not set, please set a slug manually for this scheduled command using the `sentryMonitor(\'your-monitor-slug\')` macro.');
             }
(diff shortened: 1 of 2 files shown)


src/Sentry/Laravel/Features/ConsoleSchedulingIntegration.php
```diff
@@ -266,7 +266,7 @@ private function createCheckIn(string $slug, CheckInStatus $status, ?string $id
         $options = SentrySdk::getCurrentHub()->getClient()->getOptions();
 
         return new CheckIn(
-            $slug,
+            Str::limit($slug, 128, ''),
```
joshuaruesweg, line 269: What about raising an exception here, if the `$slug` is `> 128` characters and cut of the string only, if the slug is autogenerated from a class (in `makeSlugForScheduled`)?

stayallive: We could. But that would make the code a lot more complex (to keep track where the name came from). I'm not sure there is much benefit to this. Especially since there is a very high chance the truncated string is still unique enough unless you are actively trying to break this.

cleptric (approved):

src/Sentry/Laravel/Features/ConsoleSchedulingIntegration.php
stayallive: So I remembered another reason, this is why the current only exception we throw is where it is. That if we do it here we potentially disrupt only `schedule:run` which you probably only notice after deployment instead of before.

So for now I think it's best we leave it like this and if problems arise with the limiting go the string we will revisit this and refactor to make this an exception. But I think the risk is very minimal (we didn't have any limit before and just dropped the checkin probably, so this can be seen as an improvement even).

PR #980, opened 2025-02-11
Laravel 12 support

stayallive:
~~`livewire/livewire` had to be removed because it doesn't support L12 yet so we can't have it install. We are not using it except for phpstan and in development (to have autocomplete) so this is not a critical loss and we could release without it (since dev dependencies don't end up in the release anyway).~~

Another thing is the added `"minimum-stability": "dev"` we need because L12 is not out yet. This is also just a thing for local composer installs so can be released.

Release date is planned on February 24th, so we should aim for a release of the Sentry integration a few days before that.


cleptric:
https://github.com/livewire/livewire/pull/9155 👀

cleptric (approved):

PR #992, opened 2025-05-05
Specify 'sentry-laravel' on publish tag

majidalaeinia:
It's a confusing name since there might be other things named `config` and it specifies that config name is related to the exact library.


stayallive (approved):
I've updated the name based on what other packages are doing, most packages don't even bother setting a group these days anyway and the Laravel docs also doesn't specify it and internally we don't use it. But I'm okay with updating it.

What others are called (`config` is not only Sentry here but also the Laravel Debugbar):

image

majidalaeinia:
@stayallive Can it be merged?

PR #994, opened 2025-05-13
Remove support for `traceparent`

cleptric:
Adding support for W3C's `traceparent` header was a mistake on our end. This mainly causes surprises for users, that get spammed by 3rd parties sending this header.
We made the decision to remove support from the SDK.

PR #995, opened 2025-05-13
Add back queue job traceparent propagation

cleptric:
Accidentally removed this in #994


stayallive:
Whoops, good catch @mfb!

stayallive (approved):

PR #1046, opened 2025-10-01
ci: Test against PHP 8.5

jnoordsij:
### Description

Run tests on PHP 8.5 now that first RC is available (see https://www.php.net/archive/2025.php#2025-09-25-3).

Modelled after #952.


cleptric:
Thanks a lot! We'll also need to update the base SDK. I have a wip PR https://github.com/getsentry/sentry-php/pull/1927, getting this in next week.

PR #1051, opened 2025-10-16
fix: SENTRY_LOG_LEVEL env naming

bk-ty:
### Description
Changelog states we are to use `SENTRY_LOG_LEVEL`, but the config is looking for `SENTRY_LOGS_LEVEL`, which doesn't appear to conform to the base `LOG_LEVEL` env variable.


Litarnus (commented):
Thanks for opening the PR! I added a suggestion to remain BC. It's a bit unfortunate and we will aim to remove the inconsistent option in the next major

Issue #1047, opened 2025-10-03
Logs missing from Sentry (but seen on other logging service)


danatemple:
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

My Laravel application generated the following logs, as seen on Papertrail:

```
Sep 30 03:56:03 PM UTC 169.254.3.105 php [2025-09-30 15:56:03] production.ERROR: App\Http\Controllers\ItemController::checkName unexpected error, empty name
Oct 01 03:02:48 PM UTC 169.254.9.9-1 php [2025-10-01 15:02:48] production.ERROR: Attempt to read property "id" on bool {"exception":"[object] (ErrorException(code: 0): Attempt to read property \"id\" on bool at /var/task/app/Jobs/WebDriveItemJob.php:489)
Oct 01 06:48:07 PM UTC 169.254.71.217 php [2025-10-01 18:48:07] production.ERROR: Attempt to read property "id" on bool {"userId":258047,"exception":"[object] (ErrorException(code: 0): Attempt to read property \"id\" on bool at /var/task/app/Jobs/WebDriveItemJob.php:489)
Oct 02 12:18:19 PM UTC 169.254.63.35 php  [2025-10-02 12:18:19] production.ERROR: SsoGetClassesJob::failed() {"code":"no-code"}

```

### Expected Result

To see all the logs on Sentry.

### Actual Result

However, on Sentry logging, the two lines from October 1 are missing. Here is the log list filtered by "severity = ERROR":

<img width="681" height="210" alt="Image" src="https://github.com/user-attachments/assets/98adcffd-b033-4a0a-89c6-4a53a26baea2" />

Changing the filter from "error" to e.g. looking for "attempt to read property" does not help. It seems the two log lines are just missing from Sentry.

### Product Area

Explore

### DSN

https://898f4675a355bc60e649045af6055299@o4507384124801024.ingest.us.sentry.io/4507384132009984

stayallive:
@danatemple please upgrade to https://github.com/getsentry/sentry-laravel/releases/tag/4.16.0, we previously did not log exceptions (since we already report them as exceptions) but in https://github.com/getsentry/sentry-laravel/releases/tag/4.16.0 we changed this behaviour to also log the exception message. Let me know if the issue persists when using https://github.com/getsentry/sentry-laravel/releases/tag/4.16.0 of the Laravel SDK.

danatemple:
[...]

PR #1052, opened 2025-10-20
Ensure we always flush logs after the app terminates

stayallive:
### Description

As reported in #1047, unhandled exceptions or fatal errors might not allow the logs to be flushed. This adds a terminating callback that ensures that logs are flushed in more situations, especially the one reported in #1047.

#### Issues

* resolves: #1047
* resolves: LARAVEL-35


Litarnus (approved):
:shipit:

PR #963, opened 2024-12-16
Substitute E_STRICT

cleptric:
Fixes #962

Issue #962, opened 2024-12-16
Constant E_STRICT is deprecated


jyxjjj:
### How do you use Sentry?

Self-hosted / on-premises

### SDK version

4.10.1

### Laravel version

11.35.0

### Steps to reproduce

```
php artisan sentry:test
```

### Expected result

A test issue sent to remote sentry instance

### Actual result

A deprecations error logged to local log file
```
deprecations.WARNING: ErrorException: Constant E_STRICT is deprecated in *****/vendor/sentry/sentry-laravel/src/Sentry/Laravel/Console/TestCommand.php:46
```

PR #1067, opened 2025-11-16
ref: simplify usage of E_ALL

liamduckett:
### Description
Remove indirect usage of `E_STRICT`. There is no need to use it, as `E_ALL` already contains `E_STRICT` (since [PHP 5.4.0](https://www.php.net/ChangeLog-5.php#5.4.0)). 

#### Issues
relates to: #962 (also see #963)


liamduckett:
Apologies, didn't meant to sound like any issues had resurfaced. I guess the linked issue and PR and more "related" than "relevant"?

PR #1080, opened 2026-01-12
Add log channel name to Sentry logging configuration

joshuaruesweg:
This change ensures that the logger name is logged accordingly as an attribute in Sentry. Previously, the documentation stated that this would be the case, but in practice, the logger name was never logged.


stayallive:
Im am not contesting this might be useful but for completeness I don't think I see anywhere we are supposed to be logging the channel name: https://docs.sentry.io/platforms/php/guides/laravel/logs/

What use case do you have for this specifically?

PR #1107, opened 2026-03-16
feat(ai): add invoke_agent tracing baseline

constantinius:
Add the initial AI integration wiring that records invoke_agent spans for prompting and prompted lifecycle events, including model/provider and token usage metadata. This establishes the stack base without chat, tool, message, or embeddings instrumentation.

### Included
- Registers `AiIntegration` in the feature list.
- Adds base AI tracing config flags:
  - `tracing.gen_ai`
  - `tracing.gen_ai_invoke_agent`
- Implements agent lifecycle handlers:
  - `PromptingAgent` / `StreamingAgent` -> start `invoke_agent` span
  - `AgentPrompted` / `AgentStreamed` -> finish `invoke_agent` span
- Captures baseline metadata:
  - agent name
  - request/response model
  - provider name
  - token usage
  - streaming flag
  - conversation id

Made-with: Cursor

#### Issues
- Contributes to https://linear.app/getsentry/issue/TET-2064/split-current-pr
- Contributes to https://linear.app/getsentry/issue/TET-1971/laravel-ai-integration

### DIFF
--- a/src/Sentry/Laravel/Features/AiIntegration.php
+++ b/src/Sentry/Laravel/Features/AiIntegration.php
@@ -1,0 +1,240 @@
+<?php
+
+namespace Sentry\Laravel\Features;
+
+use Illuminate\Contracts\Events\Dispatcher;
+use Sentry\SentrySdk;
+use Sentry\Tracing\Span;
+use Sentry\Tracing\SpanContext;
+use Sentry\Tracing\SpanStatus;
+
+class AiIntegration extends Feature
+{
+    private const FEATURE_KEY = 'gen_ai';
+    private const MAX_TRACKED_INVOCATIONS = 100;
+
+    /** @var array<string, AiInvocationData> */
+    private $invocations = [];
+
+    public function isApplicable(): bool
+    {
+        if (!class_exists('Laravel\Ai\Events\PromptingAgent')) {
+            return false;
+        }
+
+        return $this->isTracingFeatureEnabled(self::FEATURE_KEY);
+    }
+
+    public function onBoot(Dispatcher $events): void
+    {
+        $events->listen('Laravel\Ai\Events\PromptingAgent', [$this, 'handlePromptingAgentForTracing']);
+        $events->listen('Laravel\Ai\Events\AgentPrompted', [$this, 'handleAgentPromptedForTracing']);
+        $events->listen('Laravel\Ai\Events\StreamingAgent', [$this, 'handlePromptingAgentForTracing']);
+        $events->listen('Laravel\Ai\Events\AgentStreamed', [$this, 'handleAgentPromptedForTracing']);
+    }
+
+    public function handlePromptingAgentForTracing(\Laravel\Ai\Events\PromptingAgent $event): void
+    {
+        if (!$this->isTracingFeatureEnabled('gen_ai_invoke_agent')) {
+            return;
+        }
+
+        $parentSpan = SentrySdk::getCurrentHub()->getSpan();
+        if ($parentSpan === null || !$parentSpan->getSampled()) {
+            return;
+        }
+
+        $agentName = $this->shortClassName($event->prompt->agent);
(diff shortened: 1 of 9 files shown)


src/Sentry/Laravel/Features/AiIntegration.php
```diff
@@ -0,0 +1,240 @@
...
+    {
+        $events->listen('Laravel\Ai\Events\PromptingAgent', [$this, 'handlePromptingAgentForTracing']);
+        $events->listen('Laravel\Ai\Events\AgentPrompted', [$this, 'handleAgentPromptedForTracing']);
+        $events->listen('Laravel\Ai\Events\StreamingAgent', [$this, 'handlePromptingAgentForTracing']);
+        $events->listen('Laravel\Ai\Events\AgentStreamed', [$this, 'handleAgentPromptedForTracing']);
+    }
+
+    public function handlePromptingAgentForTracing(\Laravel\Ai\Events\PromptingAgent $event): void
+    {
+        if (!$this->isTracingFeatureEnabled('gen_ai_invoke_agent')) {
```
Litarnus: We probably should also move this string into a constant to be consistent with the code in `isApplicable`

Litarnus (commented):
We should probably also add an event listener for [AgentFailedOver](https://github.com/laravel/ai/blob/ffbb5e7666483ae18fbc0d4d0a8f25ddbdfe5541/src/Events/AgentFailedOver.php#L9).

I think without it we might end up with non finished spans in case something breaks.


Also the comment about `class_basename` seems legit, we should also prefer this over our own impl

src/Sentry/Laravel/Features/AiIntegration.php
constantinius: Addressed this using a constant.

constantinius:
> We should probably also add an event listener for [AgentFailedOver](https://github.com/laravel/ai/blob/ffbb5e7666483ae18fbc0d4d0a8f25ddbdfe5541/src/Events/AgentFailedOver.php#L9).
> 
> I think without it we might end up with non finished spans in case something breaks.

Unfortunately this Exception does not provide an invokationId, so we can't really connect it to a current run.

> Also the comment about `class_basename` seems legit, we should also prefer this over our own impl

Done

src/Sentry/Laravel/Features/AiIntegration.php
```diff
@@ -0,0 +1,233 @@
...
+
+            $oldest = $map[$oldestKey];
+            if ($oldest instanceof AiInvocationData) {
+                $oldest->span->setStatus(SpanStatus::deadlineExceeded());
+                $oldest->span->finish();
+            }
+
+            unset($map[$oldestKey]);
+        }
+    }
```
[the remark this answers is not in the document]
constantinius: evictOldestIfNeeded is a safety valve for long-running processes (Octane/queue workers) where invocations leak without a matching AgentPrompted event. It just finishes the orphaned span and
 removes it from the tracking map.

 The concern about "restoring an already-finished span" doesn't apply — the eviction code never calls setSpan(). It only calls $oldest->span->finish() and unset(). The parent span
 restoration only happens in handleAgentPromptedForTracing when the invocation completes normally, and at that point the invocation is already removed from the map, so the evicted one can't
 be looked up again.

 The scenario of "nested agent chains" where a child's eviction corrupts a parent is also not realistic at 100 concurrent in-flight invocations — that's the point of the high limit.

Litarnus:
> > We should probably also add an event listener for [AgentFailedOver](https://github.com/laravel/ai/blob/ffbb5e7666483ae18fbc0d4d0a8f25ddbdfe5541/src/Events/AgentFailedOver.php#L9).
> > I think without it we might end up with non finished spans in case something breaks.
> 
> Unfortunately this Exception does not provide an invokationId, so we can't really connect it to a current run.
> 

Ok I see, that's unlucky

Litarnus (approved):

PR #1110, opened 2026-03-16
feat(ai): capture messages, attachments, and redaction metadata

constantinius:
Add structured input/output message capture for invoke_agent and chat spans, including attachment transformation, payload truncation, and binary redaction safeguards. This layers privacy-aware message instrumentation on top of existing invoke/chat/tool tracing without introducing embeddings yet.

### Included
- Captures structured input/output messages for invoke_agent/chat flows.
- Adds attachment handling:
  - prompt attachment resolution and transformation
  - modality/type normalization (`text`, `uri`, `blob`, `file_id`, etc.)
- Adds payload size controls:
  - message truncation helpers
  - content-level truncation
- Adds binary data safeguards:
  - data-uri/base64 detection
  - blob substitution/redaction logic
- Ensures message payload attributes are only captured when `send_default_pii` is enabled.

Made-with: Cursor

#### Issues
- Contributes to https://linear.app/getsentry/issue/TET-2064/split-current-pr
- Contributes to https://linear.app/getsentry/issue/TET-1971/laravel-ai-integration

### DIFF
--- a/src/Sentry/Laravel/Features/AiIntegration.php
+++ b/src/Sentry/Laravel/Features/AiIntegration.php
@@ -1,0 +1,240 @@
+<?php
+
+namespace Sentry\Laravel\Features;
+
+use Illuminate\Contracts\Events\Dispatcher;
+use Sentry\SentrySdk;
+use Sentry\Tracing\Span;
+use Sentry\Tracing\SpanContext;
+use Sentry\Tracing\SpanStatus;
+
+class AiIntegration extends Feature
+{
+    private const FEATURE_KEY = 'gen_ai';
+    private const MAX_TRACKED_INVOCATIONS = 100;
+
+    /** @var array<string, AiInvocationData> */
+    private $invocations = [];
+
+    public function isApplicable(): bool
+    {
+        if (!class_exists('Laravel\Ai\Events\PromptingAgent')) {
+            return false;
+        }
+
+        return $this->isTracingFeatureEnabled(self::FEATURE_KEY);
+    }
+
+    public function onBoot(Dispatcher $events): void
+    {
+        $events->listen('Laravel\Ai\Events\PromptingAgent', [$this, 'handlePromptingAgentForTracing']);
+        $events->listen('Laravel\Ai\Events\AgentPrompted', [$this, 'handleAgentPromptedForTracing']);
+        $events->listen('Laravel\Ai\Events\StreamingAgent', [$this, 'handlePromptingAgentForTracing']);
+        $events->listen('Laravel\Ai\Events\AgentStreamed', [$this, 'handleAgentPromptedForTracing']);
+    }
+
+    public function handlePromptingAgentForTracing(\Laravel\Ai\Events\PromptingAgent $event): void
+    {
+        if (!$this->isTracingFeatureEnabled('gen_ai_invoke_agent')) {
+            return;
+        }
+
+        $parentSpan = SentrySdk::getCurrentHub()->getSpan();
+        if ($parentSpan === null || !$parentSpan->getSampled()) {
+            return;
+        }
+
+        $agentName = $this->shortClassName($event->prompt->agent);
+        $model = $event->prompt->model ?? null;
+        $isStreaming = is_a($event, 'Laravel\Ai\Events\StreamingAgent');
+
+        $data = [
+            'gen_ai.operation.name' => 'invoke_agent',
+            'gen_ai.agent.name' => $agentName,
+        ];
+
+        if ($isStreaming) {
+            $data['gen_ai.response.streaming'] = true;
+        }
+
+        if ($model !== null) {
+            $data['gen_ai.request.model'] = $model;
+        }
+
+        $provider = $event->prompt->provider ?? null;
+        $providerName = $provider !== null ? $provider->name() : null;
+        if ($providerName !== null) {
+            $data['gen_ai.provider.name'] = $providerName;
+        }
+
+        $temperature = $this->resolveAgentAttribute($event->prompt->agent, 'Laravel\Ai\Attributes\Temperature');
+        if ($temperature !== null) {
+            $data['gen_ai.request.temperature'] = $temperature;
+        }
+
+        $maxTokens = $this->resolveAgentAttribute($event->prompt->agent, 'Laravel\Ai\Attributes\MaxTokens');
+        if ($maxTokens !== null) {
+            $data['gen_ai.request.max_tokens'] = $maxTokens;
+        }
+
+        $agentSpan = $parentSpan->startChild(
+            SpanContext::make()
+                ->setOp('gen_ai.invoke_agent')
+                ->setData($data)
+                ->setOrigin('auto.ai.laravel')
+                ->setDescription('invoke_agent ' . ($model ?? 'unknown'))
+        );
+
+        $this->evictOldestIfNeeded($this->invocations);
+
+        $this->invocations[$event->invocationId] = new AiInvocationData($agentSpan, $parentSpan);
+        SentrySdk::getCurrentHub()->setSpan($agentSpan);
+    }
+
+    public function handleAgentPromptedForTracing(\Laravel\Ai\Events\AgentPrompted $event): void
+    {
+        $invocationId = $event->invocationId;
+        if (!isset($this->invocations[$invocationId])) {
+            return;
+        }
+
+        $invocation = $this->invocations[$invocationId];
+        $agentSpan = $invocation->span;
+        $parentSpan = $invocation->parentSpan;
+
+        $data = $agentSpan->getData();
+
+        $responseModel = $event->response->meta->model ?? null;
+        if ($responseModel !== null) {
+            $data['gen_ai.response.model'] = $responseModel;
+        }
+
+        $responseProvider = $event->response->meta->provider ?? null;
+        if ($responseProvider !== null && !isset($data['gen_ai.provider.name'])) {
+            $data['gen_ai.provider.name'] = strtolower($responseProvider);
+        }
+
+        $usage = $event->response->usage ?? null;
+        if ($usage !== null) {
+            $this->setTokenUsage($data, $usage);
+        }
+
+        $conversationId = $event->response->conversationId ?? null;
+        if ($conversationId !== null) {
+            $data['gen_ai.conversation.id'] = $conversationId;
+        }
+
+        $agentSpan->setData($data);
+        $agentSpan->setStatus(SpanStatus::ok());
+        $agentSpan->finish();
+
+        unset($this->invocations[$invocationId]);
+
+        if ($parentSpan !== null) {
+            SentrySdk::getCurrentHub()->setSpan($parentSpan);
+        }
+    }
+
+    /**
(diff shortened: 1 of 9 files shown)


src/Sentry/Laravel/Features/AiIntegration.php
```diff
@@ -501,6 +783,12 @@ private function resolveAgentAttribute(\Laravel\Ai\Contracts\Agent $agent, strin
         }
     }
 
+    private function resolveAgentInstructions(\Laravel\Ai\Contracts\Agent $agent): ?string
+    {
+        $instructions = $agent->instructions();
+        return $instructions !== null ? (string) $instructions : null;
+    }
```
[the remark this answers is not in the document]
Litarnus: Doesn't need `method_exists` check as it is part of the contract: https://github.com/laravel/ai/blob/0.x/src/Contracts/Agent.php#L17

Litarnus: I think we can inline this entirely, `Agent` has the `instructions` method and we call `$event->prompt->agent` outside anyway, so might as well change it to `$event->prompt->agent->instructions()`.

src/Sentry/Laravel/Features/AiIntegration.php
```diff
@@ -99,6 +116,23 @@ public function handlePromptingAgentForTracing(\Laravel\Ai\Events\PromptingAgent
...
+                $attachments
+            );
+            if (!empty($inputMessages)) {
+                $data['gen_ai.input.messages'] = $this->truncateMessages($inputMessages);
+            }
+
+            $instructions = $this->resolveAgentInstructions($event->prompt->agent);
+            if ($instructions !== null) {
+                $data['gen_ai.system_instructions'] = $this->truncateString($instructions);
+            }
```
Litarnus: Looks like `$instructions` can't be null here anyway, we could change it to `if (!empty($instructions))` if we want or just always add it. Might be a valuable signal to see that the system instructions were empty instead of simply missing, but not sure about this

src/Sentry/Laravel/Features/AiIntegration.php
```diff
@@ -99,6 +116,23 @@ public function handlePromptingAgentForTracing(\Laravel\Ai\Events\PromptingAgent
             $data['gen_ai.tool.definitions'] = $toolDefinitions;
         }
 
+        $attachments = $this->resolveAttachments($event->prompt);
+
+        if ($this->shouldSendDefaultPii()) {
+            $inputMessages = $this->buildUserInputMessageFromParts(
+                $event->prompt->prompt ?? null,
```
Litarnus: I think we can drop the null guard since the field always exists in this event

```suggestion
                $event->prompt->prompt,
```

src/Sentry/Laravel/Features/AiIntegration.php
```diff
@@ -616,14 +1159,22 @@ class AiInvocationMeta
...
-    public function __construct(string $agentName, ?string $providerName, ?string $model, ?string $toolDefinitions)
+    public function __construct(string $agentName, ?string $providerName, ?string $model, ?string $prompt, array $attachments, ?string $toolDefinitions)
     {
         $this->agentName = $agentName;
         $this->providerName = $providerName;
         $this->model = $model;
+        $this->prompt = $prompt;
+        $this->attachments = $attachments;
         $this->toolDefinitions = $toolDefinitions;
     }
```
Litarnus: I think the `$prompt` is typed too defensively here, we get it from `$event->prompt->prompt` which is not null

```suggestion
    /** @var string*/
    public $prompt;

    /** @var array */
    public $attachments;

/** @var string|null */
public $toolDefinitions;

    public function __construct(string $agentName, ?string $providerName, ?string $model, string $prompt, array $attachments, ?string $toolDefinitions)
{
$this->agentName = $agentName;
$this->providerName = $providerName;
$this->model = $model;
        $this->prompt = $prompt;
        $this->attachments = $attachments;
$this->toolDefinitions = $toolDefinitions;
}
```

src/Sentry/Laravel/Features/AiIntegration.php
```diff
@@ -417,6 +499,206 @@ private function resolveProviderUrlPrefix(\Laravel\Ai\Providers\Provider $provid
...
+        }
+
+        if (empty($parts)) {
+            return [];
+        }
+
+        return [
+            ['role' => 'user', 'parts' => $parts],
+        ];
+    }
```
Litarnus: `$promptText` cannot be null given the Event contracts, so we can assume that here too (see other comments)


```suggestion
    private function buildUserInputMessageFromParts(string $promptText, array $attachmentParts): array
    {
        $parts = [];

        if ($promptText !== '') {
            $parts[] = ['type' => 'text', 'content' => $promptText];
        }

        foreach ($attachmentParts as $attachmentPart) {
            $parts[] = $attachmentPart;
        }

        if (empty($parts)) {
            return [];
        }

        return [
            ['role' => 'user', 'parts' => $parts],
        ];
    }
```

src/Sentry/Laravel/Features/AiIntegration.php
```diff
@@ -417,6 +499,206 @@ private function resolveProviderUrlPrefix(\Laravel\Ai\Providers\Provider $provid
...
+
+            if ($toolResult->name !== null) {
+                $resultPart['name'] = $toolResult->name;
+            }
+
+            $messages[] = ['role' => 'tool', 'parts' => [$resultPart]];
+        }
+
+        return $messages;
+    }
```
Litarnus: This method seems overly defensive and doesn't take into account the actual return types of the contract. 
* `$source->text` is a non null string in both `TextResponse` and `Step`
* `$source->toolCalls` is a `Collection` or a not null `array` so we don't have to guard it with `?? []` 
*  If we add a check that `$toolResult` is type `ToolResult`, we can omit the null checks for `id` and `name` and always attach them. Unfortunately only `Step` gives typehints that it's a `ToolResult`, so I wouldn't fully assume that it's the same for `TextResponse`, even though it's probably the case


```suggestion
    private function buildOutputMessages($source): array
    {
        $messages = [];
        $parts = [];

        $text = $source->text;
        if ($$text !== '') {
            $parts[] = ['type' => 'text', 'content' => $text];
        }

        foreach (($source->toolCalls ?? []) as $toolCall) {
            $parts[] = $this->buildToolCallPart($toolCall);
        }

        if (!empty($parts)) {
            $messages[] = ['role' => 'assistant', 'parts' => $parts];
        }

        foreach ($source->toolResults as $toolResult) {
            if (!$toolResult instanceof \Laravel\Ai\Responses\Data\ToolResult) {
                continue;
            }
            $result = $toolResult->result;
            if ($result === null) {
                continue;
            }

            $resultContent = \is_string($result) ? $result : json_encode($result);
            if ($resultContent === false) {
                continue;
            }

            $messages[] = [
                 'role' => 'tool',
                 'parts' => [
                     'type' => 'tool_call_response',
                     'content' => $resultContent,
                     'id' => $toolResult->id,
                     'name' => $toolResult->name
            ]];
        }

        return $messages;
    }
```

src/Sentry/Laravel/Features/AiIntegration.php
```diff
@@ -417,6 +499,206 @@ private function resolveProviderUrlPrefix(\Laravel\Ai\Providers\Provider $provid
...
+        $args = $toolCall->arguments;
+        if ($args !== null) {
+            $encoded = \is_string($args) ? $args : json_encode($args);
+            if ($encoded !== false) {
+                $part['arguments'] = $encoded;
+            }
+        }
+
+        return $part;
+    }
```
Litarnus: This method can also be simplified given the contract of `ToolCall`


```suggestion
        private function buildToolCallPart(\Laravel\Ai\Responses\Data\ToolCall $toolCall): array
    {
        $part = [
            'type' => 'tool_call',
            'id' => $toolCall->id,
            'name' => $toolCall->name,
        ];
        
        $encoded = \json_encode($toolCall->arguments);
        if ($encoded !== false) {
            $part['arguments'] = $encoded;
        }
        
        return $part;
    }
```

src/Sentry/Laravel/Features/AiIntegration.php
```diff
@@ -336,37 +388,46 @@ private function findMatchingInvocation(string $url): ?string
...
         $chatSpans = $this->invocations[$invocationId]->chatSpans;
         if (empty($chatSpans)) {
             return;
         }
 
-        $steps = $response->steps ?? [];
+        /**
+         * @var $steps Step[]
+         */
+        $steps = $response->steps;
```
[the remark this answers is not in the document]
Litarnus: Steps is a collection and is not null: https://github.com/laravel/ai/blob/0.x/src/Responses/TextResponse.php#L19

src/Sentry/Laravel/Features/AiIntegration.php
```diff
@@ -112,49 +156,57 @@ public function handlePromptingAgentForTracing(\Laravel\Ai\Events\PromptingAgent
...
-        $usage = $event->response->usage ?? null;
-        if ($usage !== null) {
-            $this->setTokenUsage($data, $usage);
-        }
-
-        $conversationId = $event->response->conversationId ?? null;
-        if ($conversationId !== null) {
-            $data['gen_ai.conversation.id'] = $conversationId;
-            $this->setConversationIdOnSpans($invocationId, $conversationId);
+        $this->setTokenUsage($data, $event->response->usage);
```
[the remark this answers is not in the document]
Litarnus: Usage is not nullable: https://github.com/laravel/ai/blob/0.x/src/Responses/TextResponse.php#L21

src/Sentry/Laravel/Features/AiIntegration.php
```diff
@@ -417,25 +497,211 @@ private function resolveProviderUrlPrefix(\Laravel\Ai\Providers\Provider $provid
...
+    {
+        $messages = [];
+        $parts = [];
+
+        $text = $source->text;
+        if ($text !== '') {
+            $parts[] = ['type' => 'text', 'content' => $text];
+        }
+
+        foreach ($source->toolCalls as $toolCall) {
```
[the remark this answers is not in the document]
Litarnus: They are arrays or collections, so the property itself will not be null but instead just empty

src/Sentry/Laravel/Features/AiIntegration.php
```diff
@@ -62,25 +89,22 @@ public function handlePromptingAgentForTracing(\Laravel\Ai\Events\PromptingAgent
...
 
         $agentName = class_basename($event->prompt->agent);
-        $model = $event->prompt->model ?? null;
+        $model = $event->prompt->model;
         $isStreaming = is_a($event, 'Laravel\Ai\Events\StreamingAgent');
 
         $data = [
             'gen_ai.operation.name' => 'invoke_agent',
             'gen_ai.agent.name' => $agentName,
+            'gen_ai.request.model' => $model
```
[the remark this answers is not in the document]
Litarnus: Model cannot be null: https://github.com/laravel/ai/blob/0.x/src/Prompts/Prompt.php#L12

PR #1125, opened 2026-04-02
fix(tags): preserve context and tags set in withException(..) callback

### LINKED ISSUE #1124: Context and Tags are not sent from CLI
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

4.24.0

### Laravel version

13.2.0

### Steps to reproduce

1. Configure sentry
```php
return Application::configure(basePath: dirname(__DIR__))
    ->withExceptions(function (Exceptions $exceptions): void {
        \Sentry\configureScope(function(\Sentry\State\Scope $scope) {
            $scope->setContext('Context name', [
                'key' => 'value',
            ]);

            $scope->setTag('tag_key', 'tag_value');
        });

        \Sentry\Laravel\Integration::handles($exceptions);
    })->create();
```

2. Trigger an exception from a command

```php
namespace App\Console\Commands;

use Illuminate\Console\Command;

class SampleCommand extends Command
{
    protected $signature = 'app:sample-command';
    protected $description = 'Command description';

    public function handle()
    {
        throw new \Exception('Test');
    }
}
```

3. run `php artisan app:sample-command`

### Expected result

Contex and custom tags to be sent and displayed on Sentry 

### Actual result

Context and custom tags are missing from sentry.

Note: With the same framework configuration, if an exception is thrown during an http request, then the context and custom tags are sent/displayed on sentry

Litarnus:
Fixes an issue that causes data set on scopes to be lost because Laravel first resolves the error handler and then it resolves the `HubInterface` callback.
Because we created a fresh hub there, any data set before was lost. By copying the data from the previous scope, we can now preserve the data

### DIFF
--- a/src/Sentry/Laravel/ServiceProvider.php
+++ b/src/Sentry/Laravel/ServiceProvider.php
@@ -32,6 +32,7 @@ use Sentry\SentrySdk;
 use Sentry\Serializer\RepresentationSerializer;
 use Sentry\State\Hub;
 use Sentry\State\HubInterface;
+use Sentry\State\Scope;
 use Sentry\Tracing\TransactionMetadata;
 use Throwable;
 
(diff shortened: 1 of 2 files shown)


stayallive (approved):
Very cool fix!

PR #1128, opened 2026-04-27
Fix recursive loop when detecting the session key

### LINKED ISSUE #1127: CacheIntegration::getSessionKey() causes worker-lifetime CPU/memory runaway on web requests in some environments
# CacheIntegration::getSessionKey() causes worker-lifetime CPU/memory runaway on web requests in some environments

## Summary

In environments with a custom session driver and meaningful cache event volume, `CacheIntegration::getSessionKey()` — introduced by [#1009](<https://github.com/getsentry/sentry-laravel/issues/1009>) in 4.15.2 — causes sustained CPU and memory growth over each PHP-FPM worker's lifetime until the pool saturates and the app goes down. The `runningInConsole()` gate added in 4.19.0 (addressing [LARAVEL-41](https://linear.app/getsentry/issue/LARAVEL-41/session-started-in-order-to-track-cache-hits)) only helps the CLI variant; the web-request variant is still present in 4.25.0.

We have a clean bisect, an empirical A/B on production confirming the mechanism is in `getSessionKey()`, and a trivial workaround. What we **don't** have is a static root cause explaining *why* it specifically hangs our stack while apparently not hanging other sentry-laravel users on 4.15.2+.

## Version info

* sentry-laravel: **4.15.1** clean, **4.15.2 → 4.25.0** all reproduce the hang
* sentry/sentry: cleared across 4.14.2 → 4.16.0 (Phase A of our bisect)
* laravel/framework: 12.x
* PHP: 8.3
* SESSION_DRIVER: custom (extends `Illuminate\Session\Store`; uses `NullSessionHandler` and reads/writes directly via the `Redis` facade to share session state with a legacy CodeIgniter app)
* cache.default: redis
* SENTRY_TRACES_SAMPLE_RATE: 0.02
* SENTRY_BREADCRUMBS_CACHE_ENABLED: true (default)
* SENTRY_TRACE_REDIS_COMMANDS: false (default)

<details><summary>Bisect</summary>

Performed in-place on production (can't reproduce locally). Each step: `composer require sentry/sentry-laravel:X sentry/sentry:Y && sudo service php8.3-fpm reload`. Observed under live traffic, rolled back the moment worker saturation was detected.

| Step | sentry-laravel | sentry | Repro |
| -- | -- | -- | -- |
| A1 | 4.15.1 | 4.14.2 | no |
| A2 | 4.15.1 | 4.15.0 | no |
| A3 | 4.15.1 | 4.15.1 | no |
| A4 | 4.15.1 | 4.15.2 | no |
| A5 | 4.15.1 | 4.16.0 | no |
| B1 | **4.15.2** | 4.15.2 | **yes** |

Phase A held `sentry-laravel` at 4.15.1 and walked `sentry/sentry` forward through the full 4.14.2 → 4.16.0 range — all clean. Phase B held `sentry/sentry` at 4.15.2 (proven clean in A4) and bumped `sentry-laravel` → reproduced on the first step. The `sentry/sentry` SDK is conclusively not the culprit. The 4.15.1 → 4.15.2 diff contains exactly one behavioural commit — PR [#1009](<https://github.com/getsentry/sentry-laravel/issues/1009>) (the PHP-SDK bump in [#1028](<https://github.com/getsentry/sentry-laravel/issues/1028>) is ruled out by A4).

</details>

## Empirical confirmation that `getSessionKey()` is the mechanism

After the bisect we ran an A/B on production at **4.25.0** (latest as of this writing, 10 minor releases past the break):

1. Bump prod in-place to `sentry/sentry-laravel 4.25.0` + `sentry/sentry ^4.23`.
2. Edit `vendor/sentry/sentry-laravel/src/Sentry/Laravel/Features/CacheIntegration.php` so `getSessionKey()` unconditionally returns null:

   ```php
   private function getSessionKey(): ?string
   {
       return null;
   }
   ```
3. FPM reload. **Site stable under live traffic.**
4. Reverted the patch (restored the package default). **Site went down immediately.**
5. Re-applied the patch. **Site stable again.**

Null-ing `getSessionKey()` makes `replaceSessionKey()` / `replaceSessionKeys()` no-op (`$value === null` is always false), which is functionally identical to 4.15.1's pre-[#1009](<https://github.com/getsentry/sentry-laravel/issues/1009>) behaviour for those code paths — but keeps every other fix from 4.15.2 → 4.25.0. That A/B cleanly isolates the mechanism to `getSessionKey()`.

<details><summary>What we can't explain</summary>

We can't derive from static reading *why* `getSessionKey()` specifically hangs our stack:

* `session.store` is bound as a container singleton in `Illuminate\Session\SessionServiceProvider::registerSessionDriver()`. After first resolution it's cached. `make('session.store')` is O(1) on subsequent calls.
* `Store::getId()` returns the `$id` property — no I/O, no handler call.
* Built-in session handlers do hit storage during `read()`, but session reads happen at handler-layer and don't emit cache events. They also don't emit Redis command events unless `tracing.redis_commands=true` (off in our config).
* If "session-store resolution from inside the cache-event listener causes a loop" were the mechanism, every Laravel + Redis-session + sentry-laravel 4.15.2+ user should be seeing this. The absence of issues in this tracker matching the symptom (and [LARAVEL-41](https://linear.app/getsentry/issue/LARAVEL-41/session-started-in-order-to-track-cache-hits) being DB-only + CLI-only) suggests the interaction is environment-specific.

</details>

## Relationship to [LARAVEL-41](https://linear.app/getsentry/issue/LARAVEL-41/session-started-in-order-to-track-cache-hits)

[LARAVEL-41](https://linear.app/getsentry/issue/LARAVEL-41/session-started-in-order-to-track-cache-hits) reported the same call chain (`handleCacheEventsForBreadcrumbs` → `replaceSessionKey` → `getSessionKey` → `make('session.store')` → session-driver construction) triggering unnecessary DB connections on a database-backed session driver. The 4.19.0 fix added a `runningInConsole()` gate. The gate prevents the chain when running in console/CLI contexts but still allows it on web requests. Our failure mode is a more severe version of the same underlying coupling, in a web-request context, with a different session driver.

## Reproduction guidance

We can't ship a minimal standalone repro from production, and we've been unable to reproduce locally under synthetic load. We're happy to:

* Run instrumentation on our production deploy (e.g., `memory_get_usage()` / backtrace depth logging inside `getSessionKey()` + per-request invocation counts) if you can suggest what would be most useful.
* Apply any candidate upstream fix as a patch (via `cweagans/composer-patches`) and observe under live traffic.
* Share our (redacted) `config/session.php`, `config/cache.php`, `config/sentry.php`, and the shape of our custom session store's `read()`/`write()` methods.

## Workaround

Until there's a fix, we're running 4.25.0 with `getSessionKey()` patched to `return null;`. Visibility tradeoff: cache breadcrumbs and spans include the real key instead of the `{sessionKey}` placeholder. No other behavioural change.

cc: @soundslocke

stayallive:
### Description
When the constructor of the session store performs a cache operation a infinite loop can occur where the session store causes a cache lookup which causes a session store creation which causes a cache operation etc.

This aims to solve this by tracking if we are looking up the session key and return early to prevent looping and also optimizing the way we do by not creation the session store before the framework and instead reading it from the cookie if there is no session store yet and if the store was already resolved we just use that.

This should be both more correct, faster and safer ✨ 

#### Issues
* resolves: #1127
* resolves: LARAVEL-62

### DIFF
--- a/src/Sentry/Laravel/Features/CacheIntegration.php
+++ b/src/Sentry/Laravel/Features/CacheIntegration.php
@@ -24,6 +24,13 @@ class CacheIntegration extends Feature
     use WorksWithSpans, TracksPushedScopesAndSpans, ResolvesEventOrigin;
 
     /**
+     * Indicates whether the cache integration is currently trying to detect the session key.
+     *
+     * @var bool
+     */
+    private $isDetectingSessionKey = false;
+
+    /**
      * Indicates whether to attempt to detect the session key when running in the console.
      *
      * @internal this is mainly intended for testing purposes.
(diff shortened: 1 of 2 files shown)


Litarnus (approved):
Nice fix!

johnbacon:
@stayallive Confirmed! It seems that this does the trick. We were able to update production to `sentry/sentry-laravel:dev-fix-session-key-integration@dev` without it taking down the site as it would have before. 🎉

cc: @soundslocke @markb423

stayallive:
Nice, we'll try to get this in an official release on Monday, thanks for testing!

PR #1129, opened 2026-04-27
Rename cache operations from the session drivers

### LINKED ISSUE #1082: Filter Redis sessions from cache hit/miss rate
When using Redis for sessions, it will show up in the cache hit/miss ratio. If there are many single visitors or bots, it will report a lot of cache misses which will make the cache module not very useful for regular caching statistics.

We should not include session related informations as they can make the entire cache statistics useless

stayallive:
### Description
This renames cache operations triggered by the session driver. This removes them from the cache insights which is misleading. The detection is using the already resolved manager/drivers and doesn't trigger a driver construction itself.

#### Issues
* resolves: #1082
* resolves: LARAVEL-52


Litarnus (approved):
Nice

PR #1139, opened 2026-05-20
Update .gitattributes to exclude dev files from composer dist

williamdes:
This avoids shipping dev/tooling files in the composer dist that aren't needed at runtime.

Last `.gitattributes` update was 2aa81ac (2022-10-19, #601). The files below have been added or modified since, and are still included in the composer dist:
- `/phpstan.neon export-ignore` — last touched in 1e01c7b 2023-08-03 (#740)
- `/phpstan-baseline.neon export-ignore` — last touched in 1d1a873 2026-03-12 (#1103)
- `/.php-cs-fixer.php export-ignore` — last touched in 5a3f81b 2024-11-24 (#954)
- `/CONTRIBUTING.md export-ignore` — last touched in fbcd9bd 2024-01-23 (#839)

Background reading: https://blog.madewithlove.be/post/gitattributes/

PR #1140, opened 2026-05-22
Follow-up cleanup of .gitattributes after #1139

williamdes:
This avoids shipping dev/tooling files in the composer dist that aren't needed at runtime.

Last `.gitattributes` update was a1975a8 (2026-05-22, #1139).

### Stale entry to remove
- `/.php_cs export-ignore` - file was renamed to `.php-cs-fixer.php` in 51448d1 (2022-09-30, #576 "Upgrade php-cs-fixer dependency"). The new name is already excluded.

### Entries to add
Files added or modified since the last `.gitattributes` update, still shipped in the composer dist:
- `/AGENTS.md export-ignore` - 6.7 KB dev doc
- `/CLAUDE.md export-ignore` - 6.7 KB dev doc
- `/codecov.yml export-ignore` - CI config

Background reading: https://blog.madewithlove.be/post/gitattributes/

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,10 +4,12 @@
 /.craft.yml export-ignore
 /.gitattributes export-ignore
 /.gitignore export-ignore
-/.php_cs export-ignore
 /Makefile export-ignore
 /phpunit.xml export-ignore
 /phpstan.neon export-ignore
 /phpstan-baseline.neon export-ignore
 /.php-cs-fixer.php export-ignore
 /CONTRIBUTING.md export-ignore
+/AGENTS.md export-ignore
+/CLAUDE.md export-ignore
+/codecov.yml export-ignore


.gitattributes
```diff
@@ -4,10 +4,12 @@
...
 /.gitattributes export-ignore
 /.gitignore export-ignore
-/.php_cs export-ignore
 /Makefile export-ignore
 /phpunit.xml export-ignore
 /phpstan.neon export-ignore
 /phpstan-baseline.neon export-ignore
 /.php-cs-fixer.php export-ignore
 /CONTRIBUTING.md export-ignore
+/AGENTS.md export-ignore
```
williamdes, line 13:
```suggestion
/CHANGELOG.md export-ignore
/AGENTS.md export-ignore
```

Do you also want to exclude the changelog ?

stayallive (approved):
Thanks again 👍

.gitattributes
stayallive: I feel like the CHANGELOG.md is somewhat useful to export (especially for agents source diving). So I think we should keep that in.

PR #1157, opened 2026-08-31
feat(queue): Add optional job context to queue job events

MHIGists:
### Description

Adds an opt-in `job_context` config section that enriches events and spans captured while a queue job is processing. This makes it easier to debug failing jobs by attaching:

- `memory_usage` — memory consumption during job execution
- `queue_name` — the queue/connection identity the job ran on
- `database` — database connections queried during the job
- `execution_time` — how long the job took to run
- `attempts` — the job's attempt count
- `horizon` — Laravel Horizon metadata, when available

All measurements are disabled by default, so existing users see no change in behavior unless they explicitly opt in.

This came out of custom logic I'd written internally for debugging failing jobs — figured it was generally useful enough to contribute.

#### Issues
None

### DIFF
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -159,4 +159,25 @@ return [
         'default_integrations' => env('SENTRY_TRACE_DEFAULT_INTEGRATIONS_ENABLED', true),
     ],
 
+    // Attach extra tags and structured context scoped to the currently processing queue job
+    'job_context' => [
+        // Attach PHP memory usage at job start, at capture time, and peak usage (bytes) plus the configured memory_limit
+        'memory_usage' => env('SENTRY_JOB_CONTEXT_MEMORY_USAGE_ENABLED', false),
+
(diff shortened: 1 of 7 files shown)


Litarnus:
Hi @MHIGists, thanks for opening the PR. This generally looks useful, I will test it within the next few days to see how the data looks in Sentry and will come back to this. Appreciate the effort!

MHIGists:
> Hi @MHIGists, thanks for opening the PR. This generally looks useful, I will test it within the next few days to see how the data looks in Sentry and will come back to this. Appreciate the effort!

Hey, thank you since I's half written by Gen-AI I also did my due diligence in manually testing using an empty Laravel app with some dummy jobs but I'm sure I haven't covered everything. 

Any input would be appreciated especially around the nested synchronous jobs, the latest commit might be entirely un-needed since I usually care only for the parent jobs anyways...

PR #1159, opened 2026-09-02
fix: replace backslash with dot for feature flag names

### LINKED ISSUE #1158: Class based Pennant features results in event processing error (invalid character '\\')
### How do you use Sentry?

Self-hosted / on-premises

### SDK version

4.27.0

### Laravel version

12.67.0

### Steps to reproduce

Laravel Docs for Pennant's [class based features](https://laravel.com/framework/docs/pennant#class-based-features).

Internally, when a class based feature is used, the PHP class's fully qualified name (eg: `App\Features\NewApi`) is used as the flag name. This has some DX wins for avoiding typos in flag names, being able to pull up the definition of the feature, and the ability to have a ["before" hook](https://laravel.com/framework/docs/pennant#intercepting-feature-checks) on the feature for extra control.

1. Have a bare minimum Laravel project setup, with Pennant and Sentry added.
2. `php artisan pennant:feature NewApi`
3. Activate feature for test user.
4. Use test user to cause an error, which the Sentry integration picks up.

### Expected result

Event to be recorded without event processing errors, feature flags to be found for the event. Flag name could be altered to fit in the constraints for event processing (which the fact there are constraints for allowed flag names is seemingly undocumented, to the best I can spot in the docs), but should still reasonably traceable to the feature class.

### Actual result

Event processing errors, no feature flags found for the event.

Image

`context.flags.values.0.flag` is recorded as `null` in the JSON blob for the event.

Litarnus:
Replaces `\` with `.` so that feature flag names can be processed by relay. When using class based feature flags in Pennant, it will use the full path as name, which is unfortunately not valid as per relay.

We lack official documentation for what characters are allowed (as far as I could tell), but here is the code with allowed characters directly in relay: https://github.com/getsentry/relay/blob/master/relay-event-schema/src/protocol/contexts/flags.rs#L25

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,7 +35,10 @@ env:
       "PKSA-1q6p-sqkj-8mmj",
       "PKSA-mc58-w91n-f5gv",
       "PKSA-t21r-vtr5-3mdz",
-      "PKSA-scnn-p8mm-jbft"
+      "PKSA-scnn-p8mm-jbft",
+      "PKSA-zyf5-hrxv-hrd7",
+      "PKSA-nv44-1b4d-6gjg",
+      "PKSA-9q1p-3s19-bp1q"
     ]
   COMPOSER_ADVISORY_LARAVEL_6_SYMFONY: >-
     [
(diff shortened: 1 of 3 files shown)


AdilAzhariOmsan:
Small edge case: a feature referenced with a leading separator (`\App\Features\NewApi`) comes through here as `.App.Features.NewApi` — leading dot. `Feature::active(NewApi::class)` never produces that, but a string-referenced feature can. Running the value through `ltrim(..., '\')` before the `str_replace` would cover it.

Litarnus:
Good catch @AdilAzhariOmsan! Thank you

stayallive (approved):

PR #1165, opened 2026-09-23
fix(ai): support laravel/ai 1.0 token usage

nimah79:
### Description

laravel/ai 1.0 renamed the token usage properties:

- `Usage::$promptTokens` / `$completionTokens` → `$inputTokens` / `$outputTokens`
- `cacheReadInputTokens`, `cacheWriteInputTokens` and `reasoningTokens` moved to the new `TextUsage` subclass and are now nullable
- `EmbeddingsResponse::$tokens` → `EmbeddingsResponse::$usage->inputTokens`

`AiSpanDataBag::setTokenUsage()` and the embeddings handler read the old properties. On 1.0 this raises an `Undefined property` warning inside the laravel/ai event listeners. Laravel turns the warning into an `ErrorException`, so with the AI integration enabled (the default), every agent prompt, stream and embeddings call throws in the host app.

This PR reads whichever property the installed version provides, using `??`. That doesn't warn when a property is missing and works on PHP 7.2, so both laravel/ai 0.x and 1.x are supported.

#### Tests

- Added `test/Sentry/Features/Ai/AiSpanDataBagTest.php`. It uses no laravel/ai stub classes, so it runs on both 0.x and 1.x, and it is skipped when laravel/ai is not installed.
- Passes on laravel/ai 0.1.5 (the version CI installs) and on 1.0.0. On 1.0.0 it fails without this change.
- The existing `AiIntegrationTest` still passes on 0.1.5. It can't run on 1.0 because its stubs implement the 0.x `Agent` contract (`prompt(string $prompt, ...)`), which 1.0 widened. Testing the full integration against 1.x in CI would need those stubs updated and a matrix entry. I left that out to keep this fix small.


Litarnus (approved):
Looks good to me, thank you for the fix!

Litarnus:
@nimah79 we just released `4.28.0` which contains this fix

nimah79:
Thanks @Litarnus!

